PR #20, opened 2016-03-02
Add docker compose file

asim:


.compose.yml
```diff
@@ -0,0 +1,30 @@
+registry:
```
dancannon: Should this instance of consul only be used as a micro registry?

asim: You're right. I'll fix it.

Issue #25, opened 2016-03-11
Micro web is not working properly (tested with greeter example)


swarvanusg:
![inval](https://cloud.githubusercontent.com/assets/7374058/13703875/dde39b6a-e7be-11e5-9938-b86530e4e513.png)

The error is the data value is being passed as null, and the below is coming. I'm not sure about the below error. 

```
2016/03/11 19:24:44 [WARN] Service tag "e={\"name\":\"Say.Hello\",\"request\":{\"name\":\"Request\",\"type\":\"Request\",\"values\":[{\"name\":\"Name\",\"type\":\"string\",\"values\":null}]},\"response\":{\"name\":\"Response\",\"type\":\"Response\",\"values\":[{\"name\":\"Msg\",\"type\":\"string\",\"values\":null}]},\"metadata\":{\"stream\":\"false\"}}" will not be discoverable via DNS due to invalid characters. Valid characters include all alpha-numerics and dashes.
```

Though Micro query is working fine:

```
swarvanu@swarvanu-ThinkPad-L440 ~/ $ micro query go.micro.srv.greeter Say.Hello '{"name": "Swarvanu"}'
{
	"msg": "Hello Swarvanu"
}

```

I followed the below steps:
1> `go get github.com/micro/micro`
2> `consul agent -dev -advertise=127.0.0.1`
3> `micro --enable_stats --web_address 127.0.0.1:8585 web`
4> `swarvanu@swarvanu-ThinkPad-L440 ~/goprojects/src/github.com/micro/micro/examples/greeter/server $ go run main.go` 

Is there anything wrong, in what I'm doing ?

asim:
Hi. The warning is just that, a warning. It's occurring within consul as it attempts to map tag data as txt records for DNS but obviously that's not possible with the format. This means DNS resolution doesn't work for anything registered via micro. Consul doesn't actually support base64 encoding for DNS txt records which makes it difficult to format it in anyway that would make it valid. We use tags as a hacked method to store metadata about the the services.

It's not an issue and everything functions correctly from a micro perspective. We could potentially address this in go-micro/registry to resolve the warnings but require an elegant solution. PRs welcome.

asim:
Sorry to add. Missed the thing about values not seeming to be passed through. Will have a look in a few minutes.

asim:
PR was made https://github.com/micro/micro/pull/24 which actually fixes the RPC issue and improves the implementation slightly. That's merged.

asim:
[...]

PR #26, opened 2016-03-11
Added Service and Method Selection Dropdown in the query page of Web

swarvanusg:
Hi, 

I'm new to the micro project, and its community. Its my first pull request.

I have changed the query page. I have added Selection from Dropdown option for service and its methods. The service dropdown list is populated by all available service, where the method list is populated based on what service is currently selected.

1> At First the screen looks like:

no selection

2> The drop down list of services:

serviceselection

3> The drop down list of methods for a specific service:

methodselection

4> If a service has no methods it gets a black list:

servicewithnomethod

5>  Now use a service with ease. No need to type.

useservice

The reason the output is different from the expected is the same of issue #25 
It has nothing to do with the current changes. 

Please suggest if any change required. 

Thanks 
- Swarvanu

### DIFF
--- a/web/templates.go
+++ b/web/templates.go
@@ -79,7 +79,7 @@ var (
 					<select class="form-control" type=text name=service id=service> 
 					<option disabled selected> -- select a service -- </option>
 					{{range $key, $value := .}}
-					<option class = "list-group-item" value="{{$key.Name}}">{{$key.Name}}</option>
+					<option class = "list-group-item" value="{{$key}}">{{$key}}</option>
 					{{end}}
 					</select>
 				</ul>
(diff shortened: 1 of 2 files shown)


asim:
Thanks for the PR. I like the look of it. Only nitpick. I want the ability to query anything behind the API even if its not registered. There's a notion of internal handlers that are not advertised to the registry. Would it be possible to flip between select list and input?

swarvanusg:
Yes sure. In that case we can make the service as a select list, but for the method we could have both the option. I'll work on that and update the PR accordingly.

asim:
It would be good to be able to query services that are not registered. That seems like a bizarre idea but something in practice that is quite possible.

swarvanusg:
Thank you @asim . I think API which are not registered (or hidden) could be useful for service default operations e.g. Metric collection, log collection, or some platform specific implementation maybe for monitoring or governing the services

asim:
Having given it more thought. Lets take it as is and if anyone complains we can change later. Only nitpicks now. Please use camelCase rather than underscores for variable and make the serviceMap `map[string][]*registry.Endpoint` rather than using interface for key and deferencing the endpoint.

swarvanusg:
Updated the PR as per the discussion. And tested with the latest changes (issue #25 Fix). 

latestcommits

swarvanusg:
Hi @asim ,

Added Other Method Selection Option, for Unregistered Method Support, in case we need it in future. 
So below are the descriptions:

1> While no service and Method is selected the screen looks like:

noselection


2> Now we have an option to select "- Other" that will allow us to specify any method:

theotherselectionoption


3> While "- Other" option is not selected the "Other Method" input field is disabled:

disabledfeild


4> Example of execution of a Method Specified in the "Other Method" section:

othermethodexecution


5> Care is taken to reset and disable the "Other Method" section while we change service or method:

changeservice

please suggest if any farther changes are required

Thanks,
Swarvanu

asim:
Ok looks good.

PR #35, opened 2016-04-11
Stop reverse API proxy returning 500 when a 404 is more valid

arbarlow:
Previously, when a api request was made a to service that didn't exist, the reverse proxy would return a 500 since the director inside the proxy would have a nil url and Go's httputil reverse proxy would then error accordingly.

Code wise, I had to change the file quite a lot to get this to work, sorry about that. I couldn't think of another way to make it work. A downside to this approach is that I instantiate a new 'NewSingleHostReverseProxy' for each request, but that does seem performant enough.

Happy to discuss how to do it differently!

Websocket proxying should not be effected.

Previous behaviour..

```
curl "0.0.0.0:8080/servicethatdoesntexist" -i
```
```
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token
Access-Control-Allow-Methods: POST, GET, OPTIONS, PUT, DELETE
Date: Mon, 11 Apr 2016 18:45:55 GMT
Content-Length: 0
Content-Type: text/plain; charset=utf-8
```

New behaviour...
```
curl "0.0.0.0:8080/servicethatdoesntexist" -i
```
```
HTTP/1.1 404 Not Found
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token
Access-Control-Allow-Methods: POST, GET, OPTIONS, PUT, DELETE
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Mon, 11 Apr 2016 18:45:52 GMT
Content-Length: 19
```

### DIFF
--- a/internal/handler/proxy.go
+++ b/internal/handler/proxy.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"net/url"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -91,7 +91,7 @@ func run(ctx *cli.Context) {
 	switch ctx.GlobalString("api_handler") {
 	case "proxy":
 		log.Printf("Registering API Proxy Handler at %s", ProxyPath)
-		r.PathPrefix(ProxyPath).Handler(handler.Proxy(Namespace, false))
+		r.PathPrefix(ProxyPath).Handler(handler.Proxy(Namespace))
```
asim: Enabling/Disabling WebSockets was there for a reason. It should be disabled for the API and will be enabled for the Web Proxy when that's using this code. For some reason I assumed it was already but I guess I forgot to convert over. Please revert it to being an arg.

internal/handler/proxy.go
```diff
@@ -93,56 +127,24 @@ func isWebSocket(r *http.Request) bool {
...
-
-		s, err := next()
-		if err != nil {
-			kill()
-			return
-		}
-
-		r.URL.Host = fmt.Sprintf("%s:%d", s.Address, s.Port)
-		r.URL.Scheme = "http"
+func (p *proxy) serviceHostForRequest(r *http.Request) (host string, err error) {
```
asim: Nitpick. Not using named return values anywhere else in the Micro codebase. Please just do `return fmt.Sprintf("http://%s:%d", s.Address, s.Port), nil`

api/api.go
arbarlow: Oh I see! Is there a downside to having websocket support to backend rest/http apis?

asim: Separation of concerns basically. Leaving WebSockets for web based apps. At some point we may have the API support multiple protocols but it'll be thought out before enabling. So http, websockets, xmpp, etc.

internal/handler/proxy.go
```diff
@@ -14,21 +15,54 @@ import (
...
 type proxy struct {
-	Default  *httputil.ReverseProxy
-	Director func(r *http.Request)
+	Selector  selector.Selector
+	Namespace string
+
+	regex *regexp.Regexp
+}
+
+func Proxy(ns string) http.Handler {
```
asim: Also this is going to sound super OCD but exported functions at the bottom of the file. I have a weird habit

constants
types
variables
unexported functions
methods
exported functions

Yea I know...

api/api.go
arbarlow: Ok sure, in order to disable websockets I could add something to the proxy stuct and check that in the ServerHTTP method? Though it does seem like we get websockets for free atm!

asim: Yea just add like `wsEnabled bool` field or something of that nature. Web sockets for free? Sure but its not so much about getting stuff for free as it is about patterns of development. Side effects are not cool. While some may find it useful I'm not yet sure about the usage.

arbarlow:
@asim I think that's all the comments addressed!

asim:
LGTM

arbarlow:
:rocket:

PR #38, opened 2016-04-17
Plugin Interface

asim:
Have been thinking about plugins/middleware for Micro for a while now. This is separate to go-micro plugins. This is for the API, CLI, Sidecar, Web and Bot. It's a way of adding additional flags, commands and http middleware. Taking inspiration from vulcand but building something more specific to the way micro is built. http://vulcand.github.io/middlewares.html


plugin/plugin.go
```diff
@@ -0,0 +1,82 @@
...
+func newManager() *manager {
+	return &manager{
+		plugins: make(map[string]Plugin),
+	}
+}
+
+func (m *manager) Plugins() map[string]Plugin {
+	m.Lock()
+	defer m.Unlock()
+	return m.plugins
```
HeavyHorst: After this the Plugins-map can be directly modified from outside the plugin manager. https://play.golang.org/p/WxL0dUWYjt
Is this expected ?

asim:
Yes I'm aware that map can be modified after its returned because its actually a reference to it rather than a copy. I'm not bothered by it at the moment.

asim:
On second thought. You're right. I'm going to fix it now.

PR #44, opened 2016-05-04
Added support for client auth via TLS certificates

foobarsucka:
`--tls_client_ca_file` will enable it.

### DIFF
--- a/internal/helper/helper.go
+++ b/internal/helper/helper.go
@@ -2,7 +2,9 @@ package helper
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
+	"io/ioutil"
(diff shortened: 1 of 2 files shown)


foobarsucka:
OK, so I've verified that the changes in this pull request do in fact enable cert-based client authentication for the Micro API when the `--tls_client_ca_file` option is provided along side the `--enable_tls, --tls_key_file, --tls_cert_file` options.

My verification was done via manual testing of clients talking to the Micro API server with a) no client certs configured, b) client certs from a different CA configured, and c) client certs from the correct CA configured.

If you're wanting tests added to `helper_test.go` then I'll have to figure out how to write tests for `tls.Config` options...

Issue #41, opened 2016-04-22
Add "new" command for base service templates


asim:
In this blog post https://blog.micro.mu/2016/04/18/micro-architecture.html we've defined the architecture patterns for Micro and highlighted that we're using tooling to provide guidelines for design patterns. In keeping with this trend there should be a simple command `micro new [service]` that allows the creation of a base template for {api, srv, web} services. 

1. API comes in two types; RPC handlers with api.{Request, Response} or HTTP handlers, ideally with go-restful
2. SRV; Simple, RPC service
3. Web; Template for dashboards, also include web socket support

Templates should be runnable with no extra compilation required.

Execution:
1. `micro new foo` defaults to go.micro namespace and srv type. Name becomes go.micro.srv.foo. Directory is foo or foo-srv. Ideally foo-srv for clarity.
2. `micro new --namespace=com.example --type=web foo` settable flags on command line

asim:
Starting work on this. Report back soon.

asim:
Done.

PR #54, opened 2016-05-25
micro new [service]

asim:
In relation to https://github.com/micro/micro/issues/41. This is a first crack at basically just generating some example boiler plate template. It means not having to copy and paste crap.


new/new.go
```diff
@@ -0,0 +1,211 @@
...
+	// check if dir exists
+	if _, err := os.Stat(c.GoDir); !os.IsNotExist(err) {
+		return fmt.Errorf("%s already exists", c.GoDir)
+	}
+
+	fmt.Println("creating service", c.FQDN)
+
+	// write the files
+	for _, file := range c.Files {
+		f := path.Join(c.GoDir, file.Path)
```
HeavyHorst: to be platform independent we should use filepath.Join(c.GoDir, file.Path) to create filepaths.

new/new.go
```diff
@@ -0,0 +1,211 @@
...
+	goPath := os.Getenv("GOPATH")
+
+	// don't know GOPATH, runaway....
+	if len(goPath) == 0 {
+		fmt.Println("unknown GOPATH")
+		return
+	}
+
+	goPath = strings.Split(goPath, ":")[0]
+	goDir := path.Join(goPath, "src", path.Clean(dir))
```
HeavyHorst: here also

new/new.go
```diff
@@ -0,0 +1,211 @@
...
+	parts := strings.Split(dir, "/")
+
+	if len(alias) == 0 {
+		// set as last part
+		alias = parts[len(parts)-1]
```
HeavyHorst: alias = filepath.Base(dir) should work

asim:
Looks like I've got api, srv and web template generation examples. Might be time to merge and close.

PR #57, opened 2016-06-06
Add service whitelist for RPC calls

thcyron:
This PR adds support for whitelisting services that can be called via the API gateways's RPC endpoint.

```bash
micro --api_rpc_whitelist=com.example.service.foo,com.example.service.bar
```

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -60,6 +60,14 @@ are used to resolve an API service and method to send the query to.
 micro --api_namespace=com.example.api
 ```
 
+### Whitelist Services
+
+If you don't want all services to be accessible via the RPC endpoint, specify a list of allowed services:
+
+```bash
+micro --api_rpc_whitelist=com.example.service.foo,com.example.service.bar
+```
+
 ## Testing API
 
 ### Run the example app
--- a/api/api.go
+++ b/api/api.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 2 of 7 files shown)


asim:
Thinking about a plugin implementation rather than building directly into the binary. https://github.com/micro/go-plugins/tree/master/micro/whitelist

asim:
@thcyron feel free to play with the plugin and let me know your thoughts. It ends up being applied across the entire api as opposed to just the /rpc endpoint. This may not make sense, in which case I may need to think of a better solution or actually merge this PR.

thcyron:
I wasn't able to play with the plugin yet. For now, we use this PR to restrict access.

However, as much as I like the idea of having this as a plugin, I don't feel that performing the check in a client wrapper is the right place to do it. The client shouldn't be called in the first place if calling the service is not allowed.

asim:
I'm unsure as of yet in regard to this PR hence not having merged it. Its a very specific change to add whitelisting for RPC and I think about what other specific changes might creep up. The idea with plugins is to avoid baking that sort of thing in. Also, this only applies to the api whereas the RPC handler is used across the toolkit. 

I will give it some thought.

asim:
Looking at the latest commit it appears that everything will be blocked and only anything added to the whitelist will be allowed through. This basically breaks the existing functionality. I'm not really happy to merge this. It goes back to my point about it needing to be a plugin rather than baked into the core binary. 

This comes down to differing view points on where the management of such authorisation should occur. Obviously that's either short circuited at the HTTP layer, in the Client or by the services being called.

The flip side is if someone wanted to blacklist rather than whitelist to prevent access to a few services and not have to specify a massive list of apps to whitelist. Again down to preference in the way someone manages their apps.

I'm closing the PR. Please continue to use your version in your fork, use a plugin, client wrapper or service side auth.

PR #62, opened 2016-06-27
Allow http OPTIONS requests

Rakanixu:
Allow APIs to self describe, so clients can consume it without knowing anything previosly.

On my use case, I can generate forms depending on OPTIONS request instead of harcoded them on the clients.


asim:
OPTIONS method is regarded as an attack vector because of the fact that it describes the API. I prefer to leave it disabled.

Rakanixu:
Make sense. I wonder myself if it would be possible to allow it for APIs requiring authentication?

Closing pull request.

Issue #86, opened 2017-02-03
variable for service name


GJRTimmer:
Which variable allows me to change the name of a micro api proxy service.
Screen shot is of dev environment, but a production environment will have multiple.
I want to change this name to 'PRODUCTNAME.gateway.api' for example; which variable allows me to do that. I've tried multiple and no success so far.

See screenshot
![overview-name](https://cloud.githubusercontent.com/assets/6072255/22582395/fe94bb84-e9e6-11e6-9054-d3460c5906bf.png)

asim:
There is currently no way to change the name of the api.

"Screen shot is of dev environment, but a production environment will have multiple"

Why does this matter? You do know multiple instances of a service can be registered in Consul.

GJRTimmer:
@asim The reason I would like to change it, is because of managing the consul environment, one consul cluster can manage multiple micro-api gateways which from the point of management would be quite troublesome if every micro api shows up in consul as go.micro.api.

I really hope you can follow me.

GJRTimmer:
Let me give you a screen shot.

As seen below, 5 instance(s) for go.micro.api, but their config is different, so the actual result should be

- com.gateway.api 2(instances)
- com.backend.api 3(instances)

I really hope you would help me in creating a fix for this, I've went to the code and so far this should maybe be fixed within go-micro with a  ENV var ?

![different-api](https://cloud.githubusercontent.com/assets/6072255/22583936/ed8d14fe-e9ef-11e6-8a5b-7803e18bebb2.png)

GJRTimmer:
@asim Question, is this global variable ```server_name``` supposed to handle this ?

asim:
Issue resolved in relation to the PR.

PR #87, opened 2017-02-03
fix usage for argument --server_name / MICRO_SERVER_NAME variable

GJRTimmer:
#86

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -19,6 +19,7 @@ import (
 )
 
 var (
+	ServerName   = "go.micro.api"
 	Address      = ":8080"
 	Handler      = "api"
 	RPCPath      = "/rpc"
@@ -52,6 +53,9 @@ func (s *srv) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func run(ctx *cli.Context) {
+	if len(ctx.GlobalString("server_name")) > 0 {
+		ServerName = ctx.GlobalString("server_name")
+	}
 	if len(ctx.String("address")) > 0 {
 		Address = ctx.String("address")
 	}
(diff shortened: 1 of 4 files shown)


asim:
Please change ServerName to Name and then I will merge.

GJRTimmer:
@asim Done.

@asim Is this maybe something we should do throughout the entire micro environment ?

asim:
The build has failed. This cannot be merged yet. Please fix. https://travis-ci.org/micro/micro/builds/198005246

What do you mean "something we should do throughout the entire micro environment"?

GJRTimmer:
@asim Missed one rename, repushed with rebase.

Maybe this is something not unique to only the micro/micro project.

I'm research these awesome repositories to use them within a company project, so far I'm researching the possibilities about hardcoded namespace(s) service name(s) versus custom overrides like ```go.micro.srv.event.*```

So please don't feel offended if I'm creating several PR's over the next few weeks which addresses things like this PR.

Overall this is really an awesome ecosystem.

asim:
The naming issue is only related to micro/micro because it does not call service.Init to parse command line flags. Any regular service you see throughout the repo does not have this issue.

GJRTimmer:
Hardcoded values like https://github.com/micro/event-srv/blob/master/event/event.go#L29

asim:
Yea in those cases it may make sense. Thanks for the contribution.

GJRTimmer:
Can we somehow pass a the current name within the context so we can use a simple ```fmt.Sprintf``` for the error messages ?

asim:
Join the slack if you want to discuss things rather than in issues or PRs http://slack.micro.mu

PR #101, opened 2017-05-04
micro run thing

asim:
This adds the micro run command which fetches from source, builds and runs a service. Still a WIP.

```
# fetch, build, execute
micro run github.com/service/foo
# run service manager
micro run
# defer to service manager
micro run -x github.com/service/foo
# restart on death
micro run -r github.com/service/foo
# update source on fetch
micro run -u github.com/service/foo
# get status
micro run -s github.com/service/foo
# kill a service
micro run -k github.com/service/foo
```


k1ng440:
is it still work in progress?

asim:
This was removed. It's not currently relevant.

PR #103, opened 2017-05-18
NO_ISSUE: Added support for etcdv3 via plugins.go

alexdcox:


asim:
Sorry, no plugins will be added beyond the defaults to the base toolkit. If you want the toolkit with plugins you must build it yourself.

PR #111, opened 2017-07-22
Added health check endpoint flag

alexbilbie:
This PR adds a new flag to Micro API called `health_endpoint` which can used to configure a simple HTTP endpoint that can be used by load balancers (in my case AWS Application Load Balancer) to confirm availability.

Usage:

1. `micro api --health_endpoint=/health`
1. Visit `:8080/health` (or whatever you've configured to be the endpoint) and you can expect a HTTP 200 response

This is my first ever PR for a Golang project, please do let me know if my code can be improved in any way 👍

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -107,6 +107,23 @@ func run(ctx *cli.Context) {
 		defer st.Stop()
 	}
 
+	if len(ctx.String("health_endpoint")) > 0 {
+		he := ctx.String("health_endpoint")
+		var healthEndpoint string
+		if he[0:1] == "/" {
+			healthEndpoint = he
+		} else {
+			healthEndpoint = "/" + he
+		}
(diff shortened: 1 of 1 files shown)


asim:
Thanks for the PR unfortunately though we model micro tooling differently. Rather than baking features like this into the toolkit, it's instead extensible/pluggable so that we can build with exactly what we need. See the doc for the toolkit plugins here https://micro.mu/docs/plugins-toolkit.html.

An example plugin can be found here https://github.com/micro/examples/tree/master/plugins/prefix with more in the go-plugins repo at [go-plugins/micro](https://github.com/micro/go-plugins/tree/master/micro). 

Ideally you should create a repo, something like micro-health-plugin in your personal account. If it's a public repo then it will be picked up by the explorer https://micro.mu/explore/.

PR #132, opened 2018-02-02
api: api proxy can allow websocket forwarding

alexppxela:
Now, micro api gateway does not accept websocket from client to services.
I have added a new option "--allow_websocket" to forward external websockets to services websockets.
I think it is more simple to have these api option otherwise registering a home made plugin which does the job.

---
This change is [Reviewable](https://reviewable.io/reviews/micro/micro/132)

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -23,15 +23,16 @@ import (
 )
 
 var (
-	Name         = "go.micro.api"
-	Address      = ":8080"
-	Handler      = string(api.Default)
-	RPCPath      = "/rpc"
-	APIPath      = "/"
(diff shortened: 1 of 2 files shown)


asim:
I have to give this some thought as I'm not sure it's the right approach. Websockets were disabled in the API because it was being kept very focused on the json/proto request/response model. WebSockets are enabled in `micro web` since that's focused on web apps.

alexppxela:
Ok, let me known.

asim:
So my goal will be to include the [web](https://github.com/micro/go-api/blob/master/handler/web/web.go) handler as a configurable handler for the micro api. This will enable websocket support. Closing this PR for now.

PR #162, opened 2018-05-15
add api address

weisd:
web listen on :8082, api listen on :8080
address not found  when call service

add **api_address** flag to set api address


asim:
I can't accept this request. `/rpc` is an endpoint on micro web.

PR #166, opened 2018-05-19
Exit with status 1 on error

tudurom:


asim:
Understand the need for the change but not a fan of the approach. I will look to add something.

PR #204, opened 2018-10-10
Add: control timeout while serving HTTP for RPC.

jiusanzhou:
Access timeout of RPC by HTTP's header `Timeout`.
The default timeout is 10 second.

### DIFF
--- a/internal/handler/rpc.go
+++ b/internal/handler/rpc.go
@@ -5,12 +5,12 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/micro/go-micro/client"
 	"github.com/micro/go-micro/cmd"
 	"github.com/micro/go-micro/errors"
 	"github.com/micro/micro/internal/helper"
-	"time"
 )
 
 type rpcRequest struct {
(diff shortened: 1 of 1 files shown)


asim:
Is the default timeout for RPC actually 10 seconds though?

jiusanzhou:
> Is the default timeout for RPC actually 10 seconds though?

Sorry Asim, I was just set 10 seconds as the default timeout in my own business code.
I have changing the code with the default timeout following the client.DefaultRequestTimeout.
Plz review PR again, and decide whether you can merge the PR.

asim:
This should really be

```
       var opts []client.CallOption

	timeout, _ := strconv.Atoi(r.Header.Get("Timeout"))
        // set timeout
	if timeout > 0 {
		opts = append(opts, client.WithRequestTimeout(time.Duration(timeout) * time.Second))
	}

 	// remote call
	if len(address) > 0 {
                opts = append(opts, client.WithAddress(address))
        }

	err = (*cmd.DefaultOptions().Client).Call(ctx, req, &response, opts...)
```

internal/handler/rpc.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/micro/go-micro/cmd"
 	"github.com/micro/go-micro/errors"
 	"github.com/micro/micro/internal/helper"
+	"time"
```
asim: See the styling. This should be imported with the other stdlib imports.

PR #217, opened 2018-12-04
add http cors custom header support

lonphy:
sometimes we need custom header support at browser env(eg. xhr cros)


asim:
See this plugin https://github.com/micro/go-plugins/tree/master/micro/cors

PR #232, opened 2019-01-31
Passing cmd micro options

jiyeyuran:


jiyeyuran:
If micro supports micro options, then the external do things like:
`
opts := []micro.Option{micro.WrapClient(ratelimit.NewClientWrapper())}
cmd.Init(opts...)
`

asim:
This implies you are directly modifying the code which is meant to be a standalone binary.

jiyeyuran:
sometimes micro needs to be working as library. For example, I want to use some custom micro plugin, then the micro will be a library.
```
func main() {
	name := "gateway"

	plugin.Register(plugin.NewPlugin(
		plugin.WithName(name),
		plugin.WithHandler(func(h http.Handler) http.Handler {
			return myHandler(h)
		}),
	))

	cmd.Init()
}

```

asim:
Plugins should be registered but no one should be calling cmd.Init directly. This is not meant to be used as a library.

jiyeyuran:
But the micro enterprise is using it as library and calling cmd.Init directly. Here is the [code](https://github.com/microhq/enterprise/blob/aaaa93e66243f618d357a4b92781ee65f9f504a8/cmd/micro/main.go#L71):
```

func main() {
	// setup plugins
	plugins()

	// setup the command line
	dmc.Setup(cmd.App())

	// setup plugins
	commands(cmd.App())

	// initialise command line
	cmd.Init(
		cmd.Name(name),
		cmd.Description(description),
		cmd.Version(version),
	)
}
```

asim:
Our own internal use is controlled. We support any breaking changes we make. We will not support backwards compatibility for people using this as a library. Please do not use it as a library.

PR #241, opened 2019-03-08
Micro web New UI version 1 

printfcoder:


asim:
Thanks for this. I just got off an 11 hour flight and about to sleep so will be a while before I review but initial pass looks good.

PR #248, opened 2019-03-27
The great UI is coming

printfcoder:


asim:
I just realised this is into the v2 branch. There's no reason why I should not merge and we can iterate.

PR #250, opened 2019-04-06
 optimize display of endpoints data

printfcoder:
WeChatWorkScreenshot_dd0eedbf-c582-425d-9d29-d28961fea679

to

WeChatWorkScreenshot_02b36460-5cf4-4498-9709-74c7418b9665


asim:
Will merge. I'm not a fan of popups though. We need this to be a clean flow single window.

asim:
Other thing to note here is. Look at the way the current dashboard displays endpoints. You want it to be clean, simple and clear. We don't want to read a json blob for all endpoints. The dashboard needs to interpret these things. 

Here's a separate example a friend of mine wrote early one https://github.com/Margatroid/micro-dashboard

printfcoder:
yea, see

PR #257, opened 2019-04-23
Fix json render &  popups

printfcoder:
I can not reproduce the wrong links, so I need more test cases.


printfcoder:
ezgif com-video-to-gif

printfcoder:
ezgif com-video-to-gif (1)

PR #280, opened 2019-06-19
Server package

milosgajdos83:
This PR introduces `server` package which adds a new tool into the toolkit: `micro server`.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -16,6 +16,8 @@ var (
 	Address = ":8083"
 	// Network address to bind to
 	Network = ":9093"
+	// Router address to bind to
+	Router = ":9094"
 )
 
 func run(ctx *cli.Context, srvOpts ...micro.Option) {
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -0,0 +1,185 @@
...
+
+func Commands(options ...micro.Option) []cli.Command {
+	command := cli.Command{
+		Name:  "server",
+		Usage: "Run the micro network server",
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:   "address",
+				Usage:  "Set the micro server address :8083",
+				EnvVar: "MICRO_SERVER_ADDRESS",
```
asim: Interesting, does this class with MICRO_SERVER_ADDRESS set by go-micro?

asim:
Seems fairly straightforward. Don't really have any issues with it beyond checking if it conflicts on MICRO_SERVER_ADDRESS. Otherwise we should merge tomorrow if all builds.

PR #284, opened 2019-07-01
Added MICRO_GATEWAY_ADDRESS. Update for router.Advertise() 

milosgajdos83:
This PR adds `MICRO_GATEWAY_ADDRESS` into `micro router`.

It also accommodates for `router.Advertise()` function introduced in https://github.com/micro/go-micro/commit/c84f101e17a2bd7b4652e72f79eb758cea1a28b1

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -2,7 +2,6 @@ package router
 
 import (
 	"os"
-	"sync"
 	"time"
 
 	"github.com/micro/cli"
@@ -18,9 +17,9 @@ var (
 	Name = "go.micro.router"
 	// Address is the router microservice bind address
 	Address = ":8084"
-	// Router is the router bind address
+	// Router is the router gossip bind address
 	Router = ":9093"
-	// Network is the router network id
+	// Network is the network id
 	Network = "local"
 )
 
@@ -43,6 +42,11 @@ func run(ctx *cli.Context, srvOpts ...micro.Option) {
 	if len(ctx.String("network")) > 0 {
 		Network = ctx.String("network")
 	}
+	// default gateway address
+	var gateway string
+	if len(ctx.String("gateway")) > 0 {
+		gateway = ctx.String("gateway")
+	}
 
 	// Initialise service
 	service := micro.NewService(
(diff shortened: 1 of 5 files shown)


router/router.go
```diff
@@ -120,6 +107,11 @@ func Commands(options ...micro.Option) []cli.Command {
 				Usage:  "Set the micro network address: local",
 				EnvVar: "MICRO_NETWORK_ADDRESS",
 			},
+			cli.StringFlag{
+				Name:   "gateway",
```
asim: Should this be gateway_address and we use gateway for a name like "go.micro.network" or "go.micro.gateway" which is then looked up by the router?

milosgajdos83: Hm, sure, but should we then make the same change for `router`? i.e. to `router_address`?

asim: Those will have to change. Assumption is MICRO_{PROXY, ROUTER, NETWORK} sets the name or id and MICRO_{X}_ADDRESS sets the address. Where name is set its resolved via registry to get a list of addresses to call or its just the name needed e.g MICRO_NETWORK is the network name like local or micro.mu. Gateway would be the same. Would essentially allow dynamic resolution of gateways. 

Also Router variable is redundant since you use Address in service and the router interface itself has no address and relies on the same thing. Not sure why they differ in port but I'm assuming this is because of legacy reasons.

milosgajdos83: `Router` is bind address for ` router.Address` indeed, but `Address` is for `micro.Service.Address`

asim: But nothing is creating a separate server here right?

milosgajdos83: `router` runs as a microservice,  so besides the one created for the microservice which has the `router` handler registered on it, none other is created.

milosgajdos83: Or are you saying the `Router` [address] should be used for both the microservice server and the router itself?

asim: `Address` variable should be used for both. `Router` variable shoud be removed. Unless there will be some sort of separate internal server for the Advertise/Update side of things. 

Maybe it should just be left until we get the other stuff working. Could be that router is configured with a NewNetwork and has the internals for gossiping this.

milosgajdos83: Yeah, I think it's most likely that -> a leftover from the previous discussions when we thought of the `Router` having multiple interfaces (such as the Update/Advert one) besides the `micro.Server` one 🤔

PR #291, opened 2019-07-11
Pubsub events for micro router

milosgajdos83:
This PR makes router pub/sub to `router.Advert`s and apply the ones that were sent by other routers to the receiving router's routing table.


router/router.go
```diff
@@ -21,8 +25,127 @@ var (
...
+		route := table.Route{
+			Service: event.Route.Service,
+			Address: event.Route.Address,
+			Gateway: event.Route.Gateway,
+			Network: event.Route.Network,
+			Link:    event.Route.Link,
+			Metric:  int(event.Route.Metric),
+		}
+
+		switch action {
```
asim: This is interesting. Would you not call s.router.Process(advert) and let the internal code deal with it?

milosgajdos83: yeah that should do it. Will open a new PR shortly.

PR #304, opened 2019-08-12
Start router before advertising

milosgajdos83:
This PR depends on https://github.com/micro/go-micro/pull/663


milosgajdos83:
The test fails because `go.mod` is still pinned to the last release, which has old `Router` interface definition.

asim:
Merged the go-micro PR. The router is also included in the proxy/tunnel services so we'll need to call start there as well.

milosgajdos83:
Ah yes. Somehow I only tested the local bit and completely skipped the service one -- let's blame it on late night PR work :-)

PR #305, opened 2019-08-15
add switch to open or close /rpc for api

printfcoder:
add bool flag `--enable_rpc` for `micro api` and its default is false which means making the path `/rpc` disable.

if you want to open:

```
micro api --enable_rpc=true
```


asim:
LGTM

PR #308, opened 2019-08-22
[WIP] Network node

milosgajdos83:
This PR adds a `network` command which allows to run `micro` network node.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -21,12 +21,12 @@ import (
 var (
 	// Name of the network service
 	Name = "go.micro.network"
-	// Address is the tunnel address
+	// Address is the network address
 	Address = ":8084"
-	// Tunnel is the name of the tunnel
+	// Tunnel is the name of the network tunnel
 	Tunnel = "tun:0"
-	// Resolver is network resolver
-	Resolver = "dns"
+	// Resolver is the network resolver
+	Resolver = "registry"
 )
 
 // run runs the micro server
@@ -108,7 +108,8 @@ func run(ctx *cli.Context, srvOpts ...micro.Option) {
 	)
 
 	if err := service.Run(); err != nil {
-		log.Log("Network %s failed: %v", Name, err)
+		log.Logf("Network %s failed: %v", Name, err)
+		os.Exit(1)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


network/network.go
```diff
@@ -1 +1,158 @@
...
+	"github.com/micro/go-micro/router"
+	"github.com/micro/go-micro/tunnel"
+	"github.com/micro/go-micro/util/log"
+)
+
+var (
+	// Name of the network service
+	Name = "go.micro.network"
+	// Address is the network address
+	Address = ":8084"
```
asim: API 8080
Proxy 8081
Web 8082
Tunnel 8083
Router 8084
Network 8085

network/network.go
```diff
@@ -1 +1,158 @@
...
+	// Tunnel is the name of the network tunnel
+	Tunnel = "tun:0"
```
asim: Unneeded

network/network.go
```diff
@@ -1 +1,158 @@
...
+		p.Init(ctx)
+	}
+
+	if len(ctx.GlobalString("server_name")) > 0 {
+		Name = ctx.GlobalString("server_name")
+	}
+	if len(ctx.String("address")) > 0 {
+		Address = ctx.String("address")
+	}
+	if len(ctx.String("tunnel_id")) > 0 {
```
asim: We don't need a tunnel id since we don't use endpoints

network/network.go
```diff
@@ -1 +1,158 @@
...
+		Tunnel = ctx.String("tunnel_id")
+		// We need host:port for the Endpoint value in the proxy
+		parts := strings.Split(Tunnel, ":")
+		if len(parts) == 1 {
+			Tunnel = Tunnel + ":0"
+		}
+	}
+	var nodes []string
+	if len(ctx.String("server")) > 0 {
```
asim: Should use `nodes` here as in network nodes

network/network.go
```diff
@@ -1 +1,158 @@
...
+	case "dns":
+		res = &dns.Resolver{}
+	case "http":
+		res = &http.Resolver{}
+	case "registry":
+		res = &registry.Resolver{}
+	}
+
+	// create a tunnel
+	tun := tunnel.NewTunnel(
```
asim: We should pass Address and Nodes to Network and allow it to deal with creating the tunnel or calling tunnel.Init

milosgajdos83: We dont seem to creat tunnel in network at the moment. We said we would pass created tunnel into its constructor?

PR #327, opened 2019-09-18
Added Makefile and Dockerfile

milosgajdos83:


Makefile
```diff
@@ -0,0 +1,30 @@
+NAME=micro
+IMAGE_NAME=micro/$(NAME)
```
asim: microhq?

milosgajdos83: ah!

PR #340, opened 2019-09-21
Update health.go

xmlking:
found this bug, as I could not overwrite `healthAddress` with  `--address=0.0.0.0:8080`  flag.   
on the side note, I would prefer using `0.0.0.0` as default address. i.g.,  healthAddress = "0.0.0.0:8088"


asim:
This is not the right fix. Please change the thing we're extracting on line 27

PR #341, opened 2019-09-22
Fix address flag

xmlking:
Also changing default IP to 0.0.0.0 as k8s polling container IP.


asim:
Don't change the default IP. If you need this set it in config. We otherwise don't want I leak this port.

PR #345, opened 2019-09-24
[WIP] Registry service

milosgajdos83:
Can't be merged before following PRs are merged and `go.mod` is updated:
* https://github.com/micro/go-micro/pull/780 
* https://github.com/micro/go-micro/pull/781

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -2,10 +2,16 @@
 package registry
 
 import (
+	"context"
+	"fmt"
+	"os"
+	"sync"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/micro/cli"
 	"github.com/micro/go-micro"
+	"github.com/micro/go-micro/registry"
 	"github.com/micro/go-micro/registry/handler"
 	pb "github.com/micro/go-micro/registry/proto"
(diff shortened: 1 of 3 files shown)


registry/registry.go
```diff
@@ -17,8 +26,180 @@ var (
...
+}
+
+func (r *reg) syncRecords(nodes []string) error {
+	if len(nodes) == 0 {
+		log.Logf("[rgistry] no nodes to sync with. skipping")
+		return nil
+	}
+
+	c := pb.NewRegistryService(Name, r.client)
+	resp, err := c.ListServices(context.Background(), &pb.ListRequest{}, client.WithAddress(nodes...))
```
asim: This should probably be a new Sync rpc endpoint so that we can further control how we sync e.g adding registry id, timestamp, streaming batches. Also because if anyone switches to any other registry beneath the covers it will just return a list of names rather than the full service. 

Going to merge anyway since this works right now. Iterative steps.

milosgajdos83: It originally was but it was doing the exact same thing as `ListServices` so I decided until we need to add more stuff into `Sync()` there is no need to create a dedicated rpc call.

PR #354, opened 2019-10-04
Registry service

milosgajdos83:


network/config/kubernetes/services/micro/registry.yaml
```diff
@@ -0,0 +1,29 @@
...
+        name: micro-registry
+    spec:
+      containers:
+      - name: registry
+        env:
+        - name: MICRO_REGISTRY
+          value: "etcd"
+        - name: MICRO_REGISTRY_ADDRESS
+          value: "etcd-cluster-client"
+        args:
```
asim: Gotta add the NATS broker also

```
        - name: MICRO_BROKER
          value: "nats"
        - name: MICRO_BROKER_ADDRESS
          value: "nats-cluster"
```

milosgajdos83: Are we no longer publishing to registry?

asim: Nah it was just temporary. In DO we should use NATS. Idea being we use scalable systems underneath for these things in the cloud; etcd, nats, and whatever else. The inbuilt http/registry broker is a hack that does not scale.

PR #372, opened 2019-10-15
Enhancing micro version information

milosgajdos83:
At the moment we are hardcoding the version of micro into the source code. This is usually ok, but we can do better.

By updating `make build` task this commit allows (but does not enforce!) the following:
* adding `git tag`, `git commit` and build date into `micro` version output
* tags docker image with both the tag and the commit
* adds a `.gitignore` file that prevents committing binaries or other unnecessary files (this is almost exact copy of the .gitignore file provided by default by GitHub when creating a new `Go` project)

What do the results look like when `make build` is invoked:
```shell
$ ./micro --version
micro version v1.11.0 (2802bc4) (2019-10-15T23:50:36+0100)
```
`make  docker` tags the `micro` image with the following tags which give a glimpse of the most recent `git` tag and `git` commit sha:
```shell
micro/micro         latest              1004ee70ea61        9 seconds ago       43.1MB
micro/micro         v1.11.0-2802bc4     1004ee70ea61        9 seconds ago       43.1MB
```

If `micro` is built without invoking `make build` the extra information is not added to the version output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,32 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
+*.o
+*.a
+*.so
+
+# Folders
+_obj
+_test
+_build
+
+# Architecture specific extensions/prefixes
+*.[568vq]
+[568vq].out
+
+*.cgo1.go
+*.cgo2.c
+_cgo_defun.c
+_cgo_gotypes.go
+_cgo_export.*
+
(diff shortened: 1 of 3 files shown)


asim:
👍

asim:
We should do the same in go-micro/config/cmd and then when we docker build services we can inject the commit hash.

PR #394, opened 2019-10-29
Network Service implementation

jakexks:
For core network nodes, we want to create a full mesh. For this we need some sort of service discovery, so we will use a service that registers them in DNS and offers resolution through the `go.micro.network.dns` micro service.


network/dns/handler/dns.go
```diff
@@ -0,0 +1,65 @@
...
+	rsp.Records = providerResponse
+	return nil
+}
+
+func (d *DNS) validateMetadata(ctx context.Context) error {
+	md, ok := metadata.FromContext(ctx)
+	if !ok {
+		return errors.New("Denied: error getting request metadata")
+	}
+	token, found := md["authtoken"]
```
asim: Let's use the `Authorisation: Bearer token`  convention here. Key being Authorization and value `Bearer token`

network/dns/dns.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/micro/micro/network/dns/handler"
+	dns "github.com/micro/micro/network/dns/proto/dns"
+)
+
+// Run is the entrypoint for network/dns
+func Run(c *cli.Context) {
+	dnsService := micro.NewService(
+		micro.Name("go.micro.network.dns"),
+		micro.Version("latest"),
```
asim: No need to set version, it will automatically be set to a timestamp based on when the service is created

PR #402, opened 2019-10-29
[WIP] Auto-Update

milosgajdos83:
**DO NOT MERGE**

This PR has been opened to track the progress for `auto-update` feature of `micro`.

This is what the implementation looks like at the moment should the things stay as they are.

Like we discussed this does not look ideal -- there should be a better way.

Next steps to explore the better way:
* make `auto_update` an option of a runtime
* the `auto_update` runtime option would provide a function that would poll for updates and allow the runtime to update itself
* this will have an extra benefit of managing the polling and runtime go routines consistently within the runtime itself (right now we are having to run/manage the updater go routine separately from the runtime)


milosgajdos83:
Dam this has merge conflicts...related to 🥁🥁🥁 `go.mod`

cmd/cmd.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/micro/go-micro"
 	"github.com/micro/go-micro/config/cmd"
 	"github.com/micro/go-micro/runtime"
+	"github.com/micro/go-micro/runtime/poller"
```
asim: Probably not necessary anymore

Should import "github.com/micro/micro/internal/update"

cmd/cmd.go
```diff
@@ -298,20 +311,34 @@ func Setup(app *ccli.App, options ...micro.Option) {
...
+		// micro runtime options
+		options := []runtime.Option{
+			runtime.Type(context.GlobalString("runtime")),
+		}
+		// Use default poller for local runtime
+		if context.GlobalBool("auto_update") {
+			options = append(options, runtime.AutoUpdate(poller.DefaultPoller))
+		}
+		// create new micro runtime
+		muRuntime := runtime.NewRuntime(options...)
```
asim: I guess can now pull runtime.DefaultRuntime

cmd/cmd.go
```diff
@@ -298,20 +311,34 @@ func Setup(app *ccli.App, options ...micro.Option) {
 			"bot",      // :????
 		}
 
+		// micro runtime options
+		options := []runtime.Option{
+			runtime.Type(context.GlobalString("runtime")),
+		}
+		// Use default poller for local runtime
+		if context.GlobalBool("auto_update") {
+			options = append(options, runtime.AutoUpdate(poller.DefaultPoller))
```
asim: runtime.NewNotifier

cmd/cmd.go
milosgajdos83: This is an outdated code. Havent pushed the new code yet. Trying to figure out how to pass notifier to runtime created in `go-micro/config/cmd`, now.

cmd/cmd.go
milosgajdos83: We should use the runtime as per the passed in cli params, right. Not a default runtime. We snhould then init it with Notifier, if auto_update option has been passed in. I'm trying to figure out how all this stuff actually works and fits in together with `go-micro/config/cmd`

cmd/cmd.go
milosgajdos83: This is an outdated code.

milosgajdos83:
This will not build before https://github.com/micro/go-micro/pulls is merged in and `go.mod` is updated 😅

milosgajdos83:
Once micro/go-micro#895 is merged in, this needs to be rebased to master that points to the latest `go-micro`.

PR #421, opened 2019-11-14
add the ability to run services locally from source

asim:
This PR adds the ability to run services locally from source. It's a simple stepping stone to the runtime managing services. The experience is something as simple as:

```
micro run service --name greeter --source examples/greeter/srv/
```

This will:
- Change to examples/greeter/srv/
- Execute `go run main.go`
- Manage the lifecycle of the service
- Will watch for file changes and rebuild
- Log everything to stdout

Name is optional and only to track in memory. Source is also optional and will run from the local directory if none is specified.


xmlking:
will it also load  `plugin.g`o if exist?

PR #429, opened 2019-11-19
micro run

milosgajdos83:
**DO NOT MERGE -- CREATING THIS EARLY TO OPEN A DISCUSSION**

This PR will enable running services by simply running `micro run...` command.


runtime/runtime.go
```diff
@@ -229,13 +283,11 @@ func Flags() []cli.Flag {
...
-			Value: ".",
+			Usage: "Set the source url of the service e.g /path/to/source",
 		},
-		// TODO: change to BoolFlag
-		cli.BoolTFlag{
+		cli.BoolFlag{
 			Name:  "local",
-			Usage: "Set to run the service local",
+			Usage: "Set to run the service from local path",
 		},
```
asim: Let's also add cli.StringSlice flag to set "env" for user defined env variables passed to an app

runtime/runtime.go
```diff
@@ -13,19 +13,41 @@ import (
...
 var (
 	// Name of the runtime
 	Name = "go.micro.runtime"
 	// Address of the runtime
 	Address = ":8088"
 )
 
+func defaultEnv() []string {
+	var env []string
+	for _, evar := range []string{"MICRO_BROKER", "MICRO_BROKER_ADDRESS", "MICRO_REGISTRY", "MICRO_REGISTRY_ADDRESS"} {
```
asim: Parse out "MICRO_"  from os.Getenv() and set all those.

milosgajdos83: FYI, it's not `os.Getenv()`, it's `os.Environ()`

PR #432, opened 2019-11-21
GO111MODULE needs to be on to go get micro

elisarver:
Because micro defines specific versions in go.mod, you need to be running with modules enabled to go get. Otherwise, odd build failures occur under various dependencies that may have drifted.


asim:
Added as a newline

PR #447, opened 2019-12-01
Debug command

milosgajdos83:
Thi PR adds `micro logs` command(s) to query debug logs.


debug/debug.go
```diff
@@ -0,0 +1,192 @@
...
+		{
+			Name:  "logs",
+			Usage: "Get logs for a service",
+			Flags: Flags(),
+			Action: func(ctx *cli.Context) {
+				getLogs(ctx, options...)
+			},
+		},
+		{
+			Name:  "stats",
```
asim: This already exists here https://github.com/micro/micro/blob/master/cli/cli.go#L363

Either we need to move that in here or scope it under debug as a subcommand `micro debug stats`

milosgajdos83: Oh, I had no idea as I never used it before; I added it here as `debug` command queries the debug handler which encompasses both `logs` and `stats`.

asim: I would just strip it out then so we can merge the PR. Can be dealt with later.

PR #482, opened 2020-01-07
Switch to loadbalanced UDP for the network service

jakexks:
As Azure doesn't support multi-AZ public IP nodes in Azure AKS, we should switch to a UDP load balancer service for `micro-network`

This means that DigitalOcean deployment becomes a snowflake, so we will have to redeploy our global network with this in mind (and fix the dns for network.micro.mu by hand). However, it probably makes things easier in the long run.


network/config/kubernetes/services/micro/network.yaml
```diff
@@ -65,7 +65,7 @@ spec:
         - name: MICRO_NETWORK_NODES
           value: "network.micro.mu:30038"
```
asim: This will probably need to just strip :30038, the dns resolver will automatically add :8085

PR #488, opened 2020-01-12
[WIP] micro config

printfcoder:
`micro config` starts a config-srv  based on go-micro, which includes  below interfaces:

- Read
- Create
- Update
- Delete
- Search
- AuditLog
- Watch

## changeset

```proto
message ChangeSet {
    bytes data = 1;
    string checksum = 2;
    string format = 3;
    string source = 4;
    int64 timestamp = 5;
}
```

## database 
   pluggable and configurable, but now we only support mysql

## table 
 
**configs** where stores configuration data

```sql
create table configs
(
    id                  varchar(255) not null
        primary key,
    author              varchar(255) null,
    comment             text         null,
    timestamp           int          null,
    changeset_timestamp int          null,
    changeset_checksum  varchar(255) null,
    changeset_data      blob         null,
    changeset_format    varchar(255) null,
    changeset_source    varchar(255) null
);

create index id
    on configs (id, author);

create index timestamp
    on configs (timestamp);
```

*change_log*

```sql
create table change_log
(
    pid                 bigint auto_increment
        primary key,
    action              varchar(6)   null,
    id                  varchar(255) not null,
    path                text         null,
    author              varchar(255) null,
    comment             text         null,
    timestamp           int          null,
    changeset_timestamp int          null,
    changeset_checksum  varchar(255) null,
    changeset_data      blob         null,
    changeset_format    varchar(255) null,
    changeset_source    varchar(255) null
);

create index timestamp
    on change_log (timestamp);
```

## test

see [unit-test](https://github.com/micro/micro/compare/master...printfcoder:master#diff-ea7cd5f4aba1e04786e0e6616a1e3647)

## web dashboard
   in next version


asim:
It would be ideal if this operated using the go-micro/store interface so we could have pluggable backend storage

config/config.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/micro/go-micro"
+	"github.com/micro/go-micro/config/reader/json"
+	"github.com/micro/go-micro/util/log"
+	"github.com/micro/micro/config/db"
+	_ "github.com/micro/micro/config/db/mysql"
+	proto "github.com/micro/micro/config/proto/config"
+)
+
+var (
+	Name        = "go.micro.srv.config"
```
asim: go.micro.config

config/config.go
```diff
@@ -0,0 +1,108 @@
...
+	DatabaseURL = "root:123@(127.0.0.1:3306)/config?charset=utf8&parseTime=true&loc=Asia%2FShanghai"
+	DB          = "mysql"
+
+	PathSplitter = "/"
+	WatchTopic   = "micro.config.watch"
```
asim: "go.micro.config.events"

config/config.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/micro/go-micro/config/reader/json"
+	"github.com/micro/go-micro/util/log"
+	"github.com/micro/micro/config/db"
+	_ "github.com/micro/micro/config/db/mysql"
+	proto "github.com/micro/micro/config/proto/config"
+)
+
+var (
+	Name        = "go.micro.srv.config"
+	DatabaseURL = "root:123@(127.0.0.1:3306)/config?charset=utf8&parseTime=true&loc=Asia%2FShanghai"
```
asim: If we're sticking with defaults "loc" should probably removed

asim: Also password probably just root or something

asim:
Please work on this on your branch before making a PR for a working solution

PR #505, opened 2020-01-24
Reduce docker image by binary file size

tboerger:
To reduce the overall docker image size I have compressed the binary
size with the help of upx. I have not executed `--brute` as this results
in an unusable binary. The image size gets decreased from 50MB down to
20MB.

Beside that I'm creating an nsswitch.conf file to make sure the Go
binary properly handles name resolution, for further reference to that
you can read up on
https://giedrius.blog/2018/11/22/etc-nsswitch-conf-and-etc-hosts-woes-with-the-alpine-and-others-docker-image-and-golang/.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,14 @@
 FROM golang:1.13-alpine as builder
-RUN apk --no-cache add make git gcc libtool musl-dev
+RUN apk --no-cache add make git gcc libtool musl-dev upx
 WORKDIR /
 COPY . /
-RUN make build
+RUN make build && upx micro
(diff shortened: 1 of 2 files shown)


tboerger:
### Before

```
micro/micro                    latest              08874cd93fbd        27 hours ago        50.3MB
```

### After

```
micro/micro                    latest              c5b465d304e2        8 seconds ago       19.9MB
```

asim:
This is great! Thanks very much. We use go-release/github actions for our release binaries as well, think we should do that there?

PR #508, opened 2020-01-28
Debug service: Timestamp & historical snapshots from memory

crufter:
Removed the WIP as I have visualised the data with a graph and it seems correct, but I'm still testing this.


debug/debug.go
```diff
@@ -128,6 +128,12 @@ func Commands(options ...micro.Option) []*cli.Command {
 					EnvVars: []string{"MICRO_DEBUG_LOG"},
 					Value:   "service",
 				},
+				&cli.IntFlag{
+					Name:    "history_size",
```
asim: window

debug/debug.go
```diff
@@ -128,6 +128,12 @@ func Commands(options ...micro.Option) []*cli.Command {
...
+					Usage:   "Specifies how many seconds of stats snapshots to retain in memory",
+					EnvVars: []string{"MICRO_DEBUG_STATS_HISTORY_SIZE"},
```
asim: MICRO_DEBUG_STATS_WINDOW

PR #535, opened 2020-02-11
CLI Authentication

ben-toogood:


auth/auth.go
```diff
@@ -47,44 +55,88 @@ func run(ctx *cli.Context, srvOpts ...micro.Option) {
...
+	// Construct the JSON for the HTTP request
+	jsonStr, err := json.Marshal(map[string]string{"token": token})
+	if err != nil {
+		fmt.Println("Invalid token. Visit https://micro.mu/platform to get a token.")
+		os.Exit(1)
+	}
+	buff := bytes.NewBuffer(jsonStr)
+
+	// Execute the HTTP request
+	resp, err := http.Post("https://api.micro.mu/auth/Validate", "application/json", buff)
```
asim: Lowercase Validate

internal/config/config.go
```diff
@@ -0,0 +1,82 @@
...
+	"os"
+	"os/user"
+	"path/filepath"
+
+	"github.com/micro/go-micro/v2/config"
+	"github.com/micro/go-micro/v2/config/source/file"
+)
+
+// FileName for micro config
+const FileName = ".micro.conf"
```
asim: Let's call it `.micro` :)

PR #540, opened 2020-02-13
Exclude Auth.Validate API from Auth

ben-toogood:


platform/kubernetes/network/auth-api.yaml
```diff
@@ -55,6 +55,8 @@ spec:
           value: "etcd-cluster-client"
         - name: MICRO_AUTH
           value: "jwt"
+        - name: MICRO_AUTH_EXCLUDE
+          value: "Validate"
```
asim: Should that be Auth.Validate?

PR #542, opened 2020-02-14
Configure Platform Profile to use Athens Proxy

ben-toogood:


ben-toogood:
I tested this and it works in production.

PR #543, opened 2020-02-14
Proxy gRPC ACME

ben-toogood:
Needs testing, do not merge.


proxy/proxy.go
```diff
@@ -34,6 +41,10 @@ var (
 	Protocol = "grpc"
 	// The endpoint host to route to
 	Endpoint string
+	// ACME (Cert management)
+	ACMEProvider          = "autocert"
+	ACMEChallengeProvider = "cloudflare"
+	ACMECA                = acme.LetsEncryptStagingCA
```
asim: Must remember to flip to production CA once tested

asim:
Made some changes over the weekend to go-micro to include tls.Config as part of the acme provider so we can consistently use it across the different types of servers rather than just the net.Listener.

The config updates here should now work. I realised that the proxy includes an external server and an internal micro service. The tls is setup on the external server which we're flipping to 443 and identified by MICRO_PROXY_ADDRESS. The internal service is just like any other service, identified by the name go.micro.proxy and MICRO_SERVER_ADDRESS, I've set this to 8081.

When this PR is merged, wait for the docker builds to complete and the proxy to rollout. Then make the k8s config change for the proxy k8s service and deployment across all clusters.

`ACMECA = acme.LetsEncryptStagingCA` will need to change to the production config after testing.

All of this means proxy.micro.mu can be served on 443 via loadbalancers and where we set --platform can be configured via MICRO_PROXY_ADDRESS=proxy.micro.mu. For services running on the platform we can set MICRO_PROXY_ADDRESS=micro-proxy:8081 which points to the k8s service, which has been defined in the runtime profile in the PR.

PR #551, opened 2020-02-17
Update Auth API MICRO_AUTH_EXCLUDE to match new syntax

ben-toogood:


platform/kubernetes/network/auth-api.yaml
```diff
@@ -56,7 +56,7 @@ spec:
         - name: MICRO_AUTH
           value: "jwt"
         - name: MICRO_AUTH_EXCLUDE
-          value: "Validate"
+          value: "Auth.Validate"
```
asim: Should we be excluding both Auth.Generate and Auth.Validate?

ben-toogood: No, this was just so we can login via the CLI. This is now semi-redundant because we're exposing auth service via the proxy. Can I merge and close the PR?

asim: Yea I think so

PR #552, opened 2020-02-17
Bump go-micro version to include transport.Init bug fix

ben-toogood:


ben-toogood:
Self merging, simple go mod bump.

PR #610, opened 2020-03-11
Stop rewriting /login to /auth

ben-toogood:


asim:
Next thing we need to figure out. If you are logged in, to display Logout. lol

PR #618, opened 2020-03-12
Update for new store interface

jakexks:
This changes no semantics, so should be optimised for List calls.. But this builds!


jakexks:
Do not merge yet, as cloudflare store ignores `store.ReadPrefix`

PR #645, opened 2020-03-16
store/handler: return original error from store/service

vtolstov:
Signed-off-by: Vasiliy Tolstov <v.tolstov@unistack.org>


vtolstov:
must be merged after go-micro counterpart

PR #672, opened 2020-03-20
Update to use go-micro auth rewrite

ben-toogood:
Depends on: https://github.com/micro/go-micro/pull/1384


ben-toogood:
Approved by @asim

PR #710, opened 2020-03-27
Tidy up auth

ben-toogood:


asim:
Has a conflict so pushing separately.

PR #739, opened 2020-04-02
Store backup and restore

jakexks:
This adds the ability to create snapshots of go-micro stores, and restore snapshots to go-micro stores.


store/snapshot/snapshotter.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"github.com/micro/go-micro/v2/store"
+	"github.com/pkg/errors"
+)
+
+// Snapshotter creates snapshots of a go-micro store
+type Snapshotter interface {
+	// Init validates the Snapshotter options and returns an error if they are invalid.
+	// Init must be called before the Snapshotter is used
+	Init(opts ...SnapshotterOption) error
```
crufter, line 18: Why not have this in the constructor of a snapshotter? Custom implementations might use additional options etc anyway.

jakexks: Yeah, wasn't sure if this was the best interface

store/snapshot/snapshotter.go
```diff
@@ -0,0 +1,135 @@
...
+	"os"
+	"sync"
+	"time"
+
+	"github.com/micro/go-micro/v2/store"
+	"github.com/pkg/errors"
+)
+
+// Snapshotter creates snapshots of a go-micro store
+type Snapshotter interface {
```
asim, line 15: Call it Snapshot or Backup

store/snapshot/snapshotter.go
asim: It's fine for now. It's internal, works with our store and the usage is CLI right. So what we want is `micro store snapshot/restore`

PR #741, opened 2020-04-03
fix: update go-micro mod

crazybber:
fix: LogStream without Error()


jakexks:
I think this is done in master already

crazybber:
@jakexks  jak, a liitle question , when i upgrade go-micro mod to latest commit by `go get ....... @master` ,it will generate a version(major + timestamp+ hash) without minor and patch version ,but i found your sumbmit with a minor and a auto patch number .

PR #752, opened 2020-04-07
API / Web Namespace Configuration

ben-toogood:
Depends on https://github.com/micro/go-micro/pull/1496


api/api.go
```diff
@@ -343,10 +352,15 @@ func Commands(options ...micro.Option) []*cli.Command {
...
-				Name:    "namespace",
-				Usage:   "Set the namespace used by the API e.g. com.example.api",
+				Name:    "api_namespace",
+				Usage:   "Set the service namespace used by the API e.g. com.example.api",
 				EnvVars: []string{"MICRO_API_NAMESPACE"},
 			},
+			&cli.StringFlag{
+				Name:    "namespace",
+				Usage:   "Set the namespace used by the API e.g. micro",
+				EnvVars: []string{"MICRO_NAMESPACE"},
```
asim, line 362: This should be added as a global flag rather than in the api.

api/api.go
```diff
@@ -46,7 +47,8 @@ var (
 	RPCPath               = "/rpc"
 	APIPath               = "/"
 	ProxyPath             = "/{service:[a-zA-Z0-9]+}"
-	Namespace             = "go.micro.api"
+	Namespace             = auth.DefaultNamespace
```
asim, line 50: This should just be referenced from the global context

api/api.go
```diff
@@ -46,7 +47,8 @@ var (
...
+	APINamespace          = "go.micro.api"
```
asim, line 51: This should remain Namespace as its scoped as the api.Namespace

api/api.go
```diff
@@ -343,10 +352,15 @@ func Commands(options ...micro.Option) []*cli.Command {
 				EnvVars: []string{"MICRO_API_HANDLER"},
 			},
 			&cli.StringFlag{
-				Name:    "namespace",
-				Usage:   "Set the namespace used by the API e.g. com.example.api",
+				Name:    "api_namespace",
```
asim, line 355: Should remain namespace since its scoped as micro api --namespace

web/web.go
```diff
@@ -48,11 +48,13 @@ var (
 	Name = "go.micro.web"
 	// Default address to bind to
 	Address = ":8082"
-	// The namespace to serve
+	// The web namespace to serve
 	// Example:
 	// Namespace + /[Service]/foo/bar
 	// Host: Namespace.Service Endpoint: /foo/bar
-	Namespace = "go.micro.web"
+	WebNamespace = "go.micro.web"
```
asim, line 55: Same as api

cmd/cmd.go
```diff
@@ -171,12 +171,6 @@ func setup(app *ccli.App) {
 			EnvVars: []string{"MICRO_REPORT_USAGE"},
 			Value:   true,
 		},
-		&ccli.StringFlag{
```
asim, line 174: Would it not make sense to have this considering all core services are probably going to depend on namespace?

internal/namespace/resolver/resolver.go
```diff
@@ -0,0 +1,81 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/micro/go-micro/v2/api/resolver"
+	"github.com/micro/go-micro/v2/auth"
+	"github.com/micro/go-micro/v2/logger"
+	"golang.org/x/net/publicsuffix"
+)
+
+func NewNamespaceResolver(srvType, namespace string) resolver.NamespaceResolver {
```
asim, line 14: NewResolver

web/web.go
```diff
@@ -52,7 +53,9 @@ var (
 	// Example:
 	// Namespace + /[Service]/foo/bar
 	// Host: Namespace.Service Endpoint: /foo/bar
-	Namespace = "go.micro.web"
+	Namespace     = "go.micro"
+	Type          = "web"
+	FullNamespace = "go.micro.web"
```
asim, line 58: Would just remove this since you have Namespace + Type

cmd/cmd.go
ben-toogood: All core services will use determine the namepsace on a request by request basis.

internal/namespace/resolver/resolver.go
ben-toogood: Shall amend

web/web.go
ben-toogood: Sure, i'll amend

PR #758, opened 2020-04-08
Fixing micro server & file store, create default rule for auth

crufter:


runtime/profile/profile.go
```diff
@@ -16,6 +16,7 @@ func Server() []string {
 		"MICRO_ROUTER=service",
 		"MICRO_RUNTIME=service",
 		"MICRO_STORE=service",
+		"MICRO_STORE_BACKEND=file",
```
asim, line 19: You don't want to do this. This sets it for the services run by the runtime itself.

server/server.go
```diff
@@ -95,6 +95,9 @@ func Run(context *cli.Context) error {
 	if len(context.String("profile")) == 0 {
 		context.Set("profile", "server")
 	}
+	os.Setenv("MICRO_STORE", "file")
```
asim: MICRO_STORE_BACKEND?

crufter: No, MICRO_STORE_BACKEND only affects the store service. But actually the runtime services can't use MICRO_STORE=service because it might not have started. So we end up with scheduling problems.

server/server.go
```diff
@@ -96,12 +95,15 @@ func Run(context *cli.Context) error {
...
 		context.Set("profile", "server")
 	}
+
 	// get the network flag
 	peer := context.Bool("peer")
 
 	// pass through the environment
 	// TODO: perhaps don't do this
 	env := os.Environ()
+	env = append(env, "MICRO_STORE=file")
```
jakexks, line 105:
```suggestion
	env = append(env, "MICRO_STORE_BACKEND=file")
```

crufter: Please see my comment to Asim's question about this.

auth/handler/rules/rules.go
```diff
@@ -37,6 +38,31 @@ func (r *Rules) Init(opts ...auth.Option) {
...
+	}
+	if len(resp.GetRules()) > 0 {
+		log.Info("Rules exists. Skipping rule injection.")
+		return
+	}
+	log.Info("Generating default rules")
+	err = r.Create(context.Background(), &pb.CreateRequest{
+		Role: "*",
+		Resource: &pb.Resource{
+			Namespace: "micro",
```
asim, line 56: Namespace "*"

PR #767, opened 2020-04-09
[WIP] micro run and micro run --server works with subfolders now

crufter:
See #1510 for details


runtime/service.go
```diff
@@ -85,6 +85,20 @@ func runService(ctx *cli.Context, srvOpts ...micro.Option) {
 	name := ctx.Args().Get(0)
 	version := "latest"
 	source := ctx.String("source")
+	// Set source here correctly per flag/environment to avoid
+	// issues down the line
+	if len(source) == 0 && !ctx.Bool("platform") {
```
asim, line 90: This change breaks logic further down https://github.com/micro/micro/blob/e3b285a911961be8328490384adb95e9e0a3eb65/runtime/service.go#L142

asim: Only set it as if ctx.Bool("server") && len(source) == 0

crufter: @asim please note the !platform parts

asim: you're setting source, look further down.

```
 if ctx.Bool("platform") && len(source) == 0 { 
```

now source is no longer blank, so platform is not setting what it needs to.

PR #774, opened 2020-04-10
fix some spelling-check in micro

crazybber:
fix some spell check

add comments to eliminate warning in vscode


asim:
we leave initialise as is

PR #788, opened 2020-04-14
Adding support for multiple environments, runtime features

crufter:
To test:

```
micro server # in one terminal tab
micro env set 127.0.0.1:8081
micro run location
```

Caveat:
First git clone takes ages (about 20-30s) so the first run/update will fail.


cli/util/util.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	if len(env) == 0 {
+		fmt.Println("No env configured")
+		os.Exit(1)
+	}
+	os.Setenv("MICRO_PROXY", "service")
+	os.Setenv("MICRO_PROXY_ADDRESS", env)
+}
+
+// IsLocal returns true if we are connected to the local micro server
```
crufter, line 33: Note: This likely won't even be needed tbh, ignore this

auth/auth.go
```diff
@@ -153,13 +154,7 @@ func run(ctx *cli.Context, srvOpts ...micro.Option) {
 }
 
 func authFromContext(ctx *cli.Context) auth.Auth {
-	if ctx.Bool("platform") {
-		os.Setenv("MICRO_PROXY", "service")
-		os.Setenv("MICRO_PROXY_ADDRESS", "proxy.micro.mu:443")
-		return srvAuth.NewAuth()
-	}
-
-	return *cmd.DefaultCmd.Options().Auth
```
asim, line 162: is authFromContext used anywhere outside of the cli in any of the services? If so we may not want them to use the service implementation, we may want them to use whats set.

cli/util/util.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// SetupCommand includes things that should run for each command.
+func SetupCommand() {
+	env, err := config.Get("env")
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	if len(env) == 0 {
```
asim, line 25: Default env if no env is set?

cli/util/util.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// IsLocal returns true if we are connected to the local micro server
+func IsLocal() bool {
+	env, err := config.Get("env")
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	if len(env) == 0 {
```
asim, line 40: Again default env if no env is set?

cli/util/util.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// IsLive returns true if we are connected to the live platform
+func IsLive() bool {
+	env, err := config.Get("env")
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	if len(env) == 0 {
```
asim, line 54: Default env if no env is set?

runtime/runtime.go
```diff
@@ -164,6 +156,9 @@ func Commands(options ...micro.Option) []*cli.Command {
 			// In future we'll also have `micro run [x]` hence `micro run service` requiring "service"
 			Name:  "run",
 			Usage: RunUsage,
+			Description: `Examples:
+			micro run https://github.com/micro/services/tree/master/helloworld
```
asim, line 160: no https

runtime/service.go
```diff
@@ -13,14 +13,13 @@ import (
...
 	"github.com/micro/go-micro/v2"
-	"github.com/micro/go-micro/v2/config/cmd"
 	"github.com/micro/go-micro/v2/runtime"
 	srvRuntime "github.com/micro/go-micro/v2/runtime/service"
 )
 
 const (
 	// RunUsage message for the run command
-	RunUsage = "Required usage: micro run [service] [version] [--source github.com/micro/services]"
+	RunUsage = "Required usage: micro run [source]"
```
asim, line 22: Need to ensure run/kill/update/ps are all inline

runtime/service.go
```diff
@@ -163,25 +145,11 @@ func runService(ctx *cli.Context, srvOpts ...micro.Option) {
...
-			fmt.Println(err)
-			return
-		}
-		name = filepath.Base(path)
-	}
-
 	// run the service
 	service := &runtime.Service{
-		Name:     name,
+		Name:     "",
```
asim, line 150: why are these blank?

crufter: Empty string usage is just cruft, will nuke.
If you ask why I don't provide a name, it's because they will be extracted at the runtime service side.

runtime/service.go
crufter: +1

auth/auth.go
crufter: No, it does not seem to be used

runtime/service.go
```diff
@@ -207,22 +173,13 @@ func killService(ctx *cli.Context, srvOpts ...micro.Option) {
...
-	if ctx.Args().Len() > 1 {
-		version = ctx.Args().Get(1)
+	source, err := extractSource(ctx.Args().Get(0))
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
 	}
-
 	service := &runtime.Service{
-		Name:    name,
```
asim, line 224: Why does the Service have no name or version?

runtime/service.go
```diff
@@ -107,42 +110,11 @@ func runService(ctx *cli.Context, srvOpts ...micro.Option) {
...
 
-	// set the image from our images if its the platform
-	if ctx.Bool("platform") && len(image) == 0 {
-		if strings.HasPrefix(name, ".") || strings.HasPrefix(name, "/") {
-			fmt.Println(RunUsage)
-			return
-		}
-
-		formattedName := strings.ReplaceAll(name, "/", "-")
-		image = fmt.Sprintf("%v/%v", Image, formattedName)
```
asim, line 138: Does this piece of code still exist somewhere? Because if not we can no longer deploy to platform

cli/util/util.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/micro/go-micro/v2/util/config"
+)
+
+const (
+	localAddress = "127.0.0.1:8081"
+	liveAddress  = "proxy.micro.mu:443"
```
asim, line 15: I do not see this used anywhere

cli/helpers.go
```diff
@@ -108,6 +111,18 @@ func callService(c *cli.Context, args []string) ([]byte, error) {
 	return clic.CallService(c, args)
 }
 
+func setEnv(c *cli.Context, args []string) ([]byte, error) {
+	if len(args) == 0 || len(args[0]) == 0 {
+		return nil, errors.New("Please supply a value")
+	}
+	return nil, config.Set(args[0], "env")
```
asim, line 118: Should this be 

`micro env set foobar foobar.micro.mu:443`

where it becomes

`
{
  "env": {
    "foobar": "foobar.micro.mu:443"
  }
}
`

runtime/service.go
crufter: Because it will be extracted by the runtime service on the backend.

asim: It will not. You need to ensure the values defined in the fields are actually sent through. Runtime is expecting Name/Version/Source. The things as options are option. These fields should be set.

cli/helpers.go
crufter: +1, I don't get your last question though.

runtime/service.go
crufter: Fixed

cli/helpers.go
crufter: Let's leave changing this to an other PR, it's a very experimental subcommand anyway, let's figure out usage by using it.

cli/util/util.go
crufter: Now it's used.

runtime/service.go
crufter: Just reintroduced.

PR #792, opened 2020-04-15
Add Priority to Rules

ben-toogood:


auth/auth.go
```diff
@@ -68,6 +68,11 @@ var (
 			Usage: "The access level, must be granted or denied",
 			Value: "granted",
 		},
+		&cli.IntFlag{
+			Name:  "priority",
+			Usage: "The priority level, default is 0, the greater the number the higher the priority",
```
asim, line 73: Should this be the other way around?

ben-toogood: It's easier to go higher since you don't know how many levels of priority will be used. Also, the default being zero is a lot easier than some arbitrary number.

asim: Usually its like whats the #1 priority rather than whats priority 10 or 100. 

If I saw the rules

grant rule platform admin 10
grant rule platform developer 1
grant rule platform user 0

I'd basically assume user has the highest priority. I think 0 being no priority makes sense. And 1 being highest.

asim: On second thought. Let's try it and see.

PR #841, opened 2020-05-01
Update to use service.Options().Store

asim:


crufter:
I have seen that you restarted the tests. The config test must be adjusted because of the "error listing rules" comes up as an error log when the auth service is not available yet.

PR #848, opened 2020-05-07
WIP: Add store handler changes

asim:
Requires merging https://github.com/micro/go-micro/pull/1604


service/store/handler/handler.go
```diff
@@ -114,10 +116,18 @@ func (s *Store) Read(ctx context.Context, req *pb.ReadRequest, rsp *pb.ReadRespo
 	}
 
 	for _, val := range vals {
+		metadata := make(map[string]*pb.Field)
+		for k, v := range val.Metadata {
+			metadata[k] = &pb.Field{
+				Type:  reflect.TypeOf(v).String(),
+				Value: fmt.Sprintf("%v", v),
```
domwong, line 123: This is a bit sketchy, if v contains a pointer field then you'll get a pointer rather than anything useful. You might need to go the long way around and encode to json to or something.

asim: Hmm, trying to think about the best way. Can you lend a hand?

domwong: I think you / we need to write up some test cases for this.  

In `write()` you're recording metadata with as mapping of `key->Field.value` which is essentially `string->string` but you drop the `Field.type` . 

In `read()` you're reading that record back by trying to inspect the type using reflect. I'm not sure it'll pick up the right thing here e.g. if you persisted an int `200` would it come back as an int or would it be a string?

asim: Yea I see this more as a placeholder that then leads to a more complete solution. Right now I expect all things will likely get marked as string. It will potentially require some form of type checking.

PR #860, opened 2020-05-13
Registry: Improved multi-tenancy support

ben-toogood:


cmd/cmd.go
```diff
@@ -282,11 +282,14 @@ func setup(app *ccli.App) {
...
 		}
 
-		// add the system rules
-		for role, resources := range inauth.SystemRules {
-			for _, res := range resources {
-				if err := (*cmd.DefaultCmd.Options().Auth).Grant(role, res); err != nil {
-					return err
+		// add the system rules if we're using the JWT implementation
+		// which doesn't have access to the rules in the auth service
+		if (*cmd.DefaultCmd.Options().Auth).String() == "jwt" {
```
ben-toogood, line 287: I had to make this change because API / Web etc which use auth service were creating the rules there also.

PR #880, opened 2020-05-19
Refactor runtime/manager with added support for mutli-tenancy

ben-toogood:


service/runtime/manager/manager.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/micro/go-micro/v2/config/cmd"
+	"github.com/micro/go-micro/v2/runtime"
+	"github.com/micro/go-micro/v2/store"
+	"github.com/micro/go-micro/v2/store/memory"
+	"github.com/micro/micro/v2/internal/namespace"
+)
+
+// Init initializes the runtime
+func (m *manager) Init(...runtime.Option) error {
+	return nil
```
domwong, line 13: Ignoring options on purpose?

service/runtime/manager/manager.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, srv := range srvs {
+		md, ok := statuses[srv.Name+":"+srv.Version]
+		if !ok {
+			continue
+		}
+		srv.Metadata["status"] = md.Status
+		srv.Metadata["error"] = md.Error
```
domwong, line 75: Can break here since you've found it

service/runtime/manager/manager.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+
+	// publish the event which will trigger a delete in the runtime
+	return m.publishEvent(runtime.Delete, srv, &runtime.CreateOptions{Namespace: options.Namespace})
+}
+
+// Starts the manager
+func (m *manager) Start() error {
+	if m.running {
```
domwong, line 128: Should we be protecting m.running with a mutex?

service/runtime/manager/services.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return m.options.Store.Write(&store.Record{Key: s.Key(), Value: bytes})
+}
+
+// readServices returns all the services in a given namespace. If a service name and
+// version are provided it will filter using these as well
+func (m *manager) readServices(namespace string, srv *runtime.Service) ([]*runtime.Service, error) {
+	prefix := servicePrefix + namespace + ":"
+	if len(srv.Name) > 0 {
```
domwong, line 45: Might want to do a check that `Name` and `Version` are both set because if `Name` isn't but `Version` is it'll mess up the prefix

service/runtime/manager/events.go
```diff
@@ -0,0 +1,168 @@
...
+			runtime.WithCommand(ev.Options.Command...),
+			runtime.WithEnv(m.runtimeEnv(ev.Options)),
+		)
+	}
+
+	// if there was an error update the status in the cache
+	if err != nil {
+		logger.Warnf("Error procesing %v event for service %v:%v in namespace %v: %v,", ev.Type, ev.Service.Name, ev.Service.Version, ns, err)
+		ev.Service.Metadata = map[string]string{"status": "error", "error": err.Error()}
+		m.cacheStatus(ns, ev.Service)
```
domwong, line 126: Should we also update the cache for non error? Rather than wait for the status watcher loop?

service/runtime/runtime.go
```diff
@@ -48,13 +49,8 @@ func Run(ctx *cli.Context, srvOpts ...micro.Option) {
 		muRuntime.Init(runtime.WithSource(ctx.String("source")))
 	}
 
-	// use default store
-	muStore := *cmd.DefaultCmd.Options().Store
```
domwong, line 52: This is a breaking change

service/runtime/manager/manager.go
ben-toogood: Yes, since the runtime we're managing is passed in on New, we don't have any use for the the options since the manager isn't actually a runtime.

service/runtime/manager/manager.go
ben-toogood: Yes, this is probably a good idea. I'm not actually sure if m.running is even necessary though.

service/runtime/manager/services.go
ben-toogood: Yes, I noticed this but i didn't think that edge-case was likely to come up. I'll amend it.

service/runtime/manager/events.go
ben-toogood: Good idea.

service/runtime/runtime.go
ben-toogood: cmd.DefaultCmd.Options().Store is used as the default when I parse the options so it won't be a breaking change.

domwong: Hm. When I build this branch and run `micro run .` on my service I get 

```
Unsupported store: file
exit status 1
```

ben-toogood: I'll swap to using `service.Options().Store()`

PR #891, opened 2020-05-22
Auth mutli-tenancy

ben-toogood:
Update auth rules and accounts to be multi-tenant. Replace roles with scopes and introduce auth account issuer.


ben-toogood:
Depends on: https://github.com/micro/go-micro/pull/1654

PR #901, opened 2020-05-27
Fixing micro new

crufter:
Pinning deps to a certain version, see go.mod template changes.


crufter:
Closing as we don't want to specify go.mod

PR #906, opened 2020-05-29
Fix default account generation and its test

crufter:
There are a bunch of service accounts already by the time the `setupDefaultAccount` runs.


service/auth/handler/auth/auth.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/micro/go-micro/v2/errors"
 	"github.com/micro/go-micro/v2/store"
 	memStore "github.com/micro/go-micro/v2/store/memory"
+	"github.com/micro/go-micro/v2/util/log"
```
ben-toogood, line 19: Please change to go-micro/v2/logger

crufter: Ehh, VS code autocomplete gets me every time

PR #907, opened 2020-05-29
Fixing the new command

crufter:
TestNew now passes https://github.com/micro/micro/pull/907/checks?check_run_id=720270995#step:7:98

### DIFF
--- a/internal/template/module.go
+++ b/internal/template/module.go
@@ -5,6 +5,8 @@ var (
 
 go 1.13
 
+// This can be removed once etcd becomes go gettable, version 3.4 and 3.5 is not,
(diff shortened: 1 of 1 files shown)


internal/template/module.go
```diff
@@ -1,5 +1,10 @@
 package template
 
 var (
-	Module = `module {{.Dir}}`
+	Module = `module {{.Dir}}
+
+go 1.13
+
```
crufter, line 7: Should add this comment perhaps?

```go
// This can be removed once etcd becomes go gettable, version 3.4 and 3.5 is not,
// see https://github.com/etcd-io/etcd/issues/11154 and https://github.com/etcd-io/etcd/issues/11931.
```

domwong: Yeah worth just chucking that in

PR #919, opened 2020-06-05
Remove redundant init block

refs:


domwong:
Not sure why this isn't showing up in the diff here but we've already fixed this so we no longer have the dupe. https://github.com/micro/micro/blob/master/service/auth/auth.go#L91-L104

Thanks for submission!

PR #930, opened 2020-06-09
Test and config cli fixes

crufter:


service/config/handler/handler.go
```diff
@@ -273,7 +273,10 @@ func (c *Config) Delete(ctx context.Context, req *pb.DeleteRequest, rsp *pb.Dele
 	// Get the current change set
 	records, err := c.Store.Read(namespace)
 	if err != nil {
-		return errors.BadRequest("go.micro.config.Update", "read old value error: %v", err)
+		if err.Error() != "not found" {
+			return errors.BadRequest("go.micro.srv.Delete", "read old value error: %v", err)
+		}
+		return nil
```
crufter, line 279: Deleting nonexistent things returned an error without this, this makes it idempotent.

service/config/config.go
```diff
@@ -115,8 +115,9 @@ func getConfig(ctx *cli.Context) error {
 		// The actual key for the val
 		Path: key,
 	})
+
 	if err != nil {
-		if strings.Contains(err.Error(), "not found") {
+		if strings.Contains(strings.ToLower(err.Error()), "not found") {
```
crufter, line 120: The error output has been changed from `not found` to `Not found`...

service/config/handler/handler.go
domwong: hm. I'm slightly on the fence about this change. On balance, I'll allow it because there are plenty of other APIs that if you delete a non-existent key you get no error (see go maps for example)

PR #931, opened 2020-06-09
Do not set namespace for non server and platform envs

crufter:
Fixes running services in the tests, see `$ go clean -testcache && go test --tags=integration  -failfast -v -run  TestStreamLogsAndThirdPartyRepo
` before and after.


crufter:
Very.

PR #933, opened 2020-06-09
Show help if subcommand is not recognised rather than run the core service

domwong:
Only done for services that have subcommands
- auth
- config
- network
- registry
- store

Addresses https://github.com/micro/development/issues/198


service/auth/auth.go
```diff
@@ -90,11 +90,6 @@ var (
 func Run(ctx *cli.Context, srvOpts ...micro.Option) {
 	log.Init(log.WithFields(map[string]interface{}{"service": "auth"}))
 
-	// Init plugins
-	for _, p := range Plugins() {
-		p.Init(ctx)
-	}
-
```
domwong, line 97: This was a dupe, plugins also inited a few lines below

PR #945, opened 2020-06-15
Fixing the new command

crufter:
Locally the `TestNew` test failed for me and this change fixed that.
However what I don't understand is how it passed in CI...

### DIFF
--- a/client/cli/util/util.go
+++ b/client/cli/util/util.go
@@ -37,25 +37,21 @@ const (
 )
 
 var defaultEnvs = map[string]Env{
-	EnvLocal: Env{
+	EnvLocal: {
 		Name:         EnvLocal,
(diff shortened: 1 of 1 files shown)


client/cli/util/util.go
```diff
@@ -66,6 +62,10 @@ func isBuiltinService(command string) bool {
 
 // SetupCommand includes things that should run for each command.
 func SetupCommand(ctx *ccli.Context) {
+	switch ctx.Args().First() {
+	case "new", "server", "help":
+		return
+	}
 	if ctx.Args().Len() == 1 && isBuiltinService(ctx.Args().First()) {
```
crufter, line 69: Because of this `ctx.Args().Len() == 1` the `micro new [name]` was not skipped as it has `Args().Len()` == 2

PR #948, opened 2020-06-15
router: exit process when an error occurs

ben-toogood:
When an error occurred with the router, the process never exited because the WaitGroup was waiting for both go-routines to complete. Since we're waiting for an error to be sent on the errChan, there is no need for the WaitGroup.

To verify the fix, run `micro router` without `micro network` (the network is used to advertise events). It should log the error and exit the process.

### DIFF
--- a/service/router/router.go
+++ b/service/router/router.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"sync"
 	"time"
 
 	"github.com/micro/cli/v2"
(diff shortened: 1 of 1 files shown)


domwong:
I feel like there's something missing from the original implementation. Normally I'd expect to see this waitgroup because there's some processing that needs to happen before exit. So you'd have both `rtr.PublishAdverts` loop and `service.Run()` loops listening on some exit channel and cleaning up before returning from their functions. Looking at the code I can't see anything like that so I'd agree that it appears that the waitgroup is not doing anything useful right now

PR #953, opened 2020-06-17
Pass through args from micro server to core services

domwong:
When running `micro server` env vars are passed to core services but global flags are not. 

e.g. `MICRO_STORE=cockroach micro server` works but `micro --store cockroach server` doesn't.

Closes https://github.com/micro/development/issues/225

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -99,7 +99,9 @@ func Run(context *cli.Context) error {
 	peer := context.Bool("peer")
 
 	// pass through the environment
-	env := []string{}
+	// By default we want a file store when we run micro server.
+	// This will get overridden if user has set their own MICRO_STORE env var or passed in --store
+	env := []string{"MICRO_STORE=file"}
 	env = append(env, "MICRO_RUNTIME_PROFILE="+context.String("profile"))
 	env = append(env, os.Environ()...)
 
--- a/server/server.go
+++ b/server/server.go
@@ -99,8 +99,7 @@ func Run(context *cli.Context) error {
 	peer := context.Bool("peer")
 
 	// pass through the environment
-	// TODO: perhaps don't do this
-	env := []string{"MICRO_STORE=file"}
+	env := []string{}
 	env = append(env, "MICRO_RUNTIME_PROFILE="+context.String("profile"))
 	env = append(env, os.Environ()...)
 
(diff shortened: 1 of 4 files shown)


ben-toogood:
Was there a reason MICRO_STORE=file was previously set and no longer required?

domwong:
> Was there a reason MICRO_STORE=file was previously set and no longer required?

Good question. It's a fairly recent change and Asim didn't really specify why https://github.com/micro/micro/pull/884

Let me have a bit more of a play around with this as precedence of urfave/cli should have meant I didn't need to remove it but I wasn't seeing that behaviour.

asim:
Store was set to file because @crufter and I decided persistence was necessary by default since accounts, data and things were just being blown away on restarts since it was the memory store before. If its the memory store, which is the default none of this works across restarts which can be problematic.

domwong:
OK no problem. Have reinstated and tested that it works as expected when passing `MICRO_STORE` env var and `--store` args. Have also added a note to the code to explain the reasoning

domwong:
Converted back to draft as I explore some further CLI weirdness with passing args

server/server.go
```diff
@@ -90,18 +90,15 @@ func Run(context *cli.Context) error {
 		cli.ShowSubcommandHelp(context)
 		os.Exit(1)
 	}
-	// set default profile
-	if len(context.String("profile")) == 0 {
-		context.Set("profile", "server")
-	}
```
domwong, line 96: Removing this because it's setting "profile" at the wrong context level. Setting at this level means you'd override the global "profile" which  is actually the debug profiling and is either http or pprof

server/server.go
```diff
@@ -90,18 +90,15 @@ func Run(context *cli.Context) error {
...
 	// get the network flag
 	peer := context.Bool("peer")
 
 	// pass through the environment
-	// TODO: perhaps don't do this
+	// By default we want a file store when we run micro server.
+	// This will get overridden if user has set their own MICRO_STORE env var or passed in --store
 	env := []string{"MICRO_STORE=file"}
-	env = append(env, "MICRO_RUNTIME_PROFILE="+context.String("profile"))
+	env = append(env, "MICRO_RUNTIME_PROFILE=server")
```
crufter, line 101: ouch...

PR #956, opened 2020-06-19
Upoad whole repo when doing `micro run` to support services with dependencies inside that repo

crufter:
This issue was triggered by the fact that while `micro run payments/provider/stripe` works, the service can't be run locally because it depends on a package in a [parent directory](https://github.com/micro/services/blob/master/payments/provider/stripe/go.mod#L12).

This is related to this ticket: https://github.com/micro/development/issues/208

While this does not fix all dependencies at least it does fix the case when a package depends on a package in the same (mono)repo.

- [x] Adding test
- [x] Implementation
- [x] Speed up
- [x] Waiting for go micro to pass tests before merging


test/runtime_test.go
```diff
@@ -490,3 +490,34 @@ func replaceStringInFile(t *t, filepath string, original, newone string) {
...
+	}
+
+	try("Find hello world", t, func() ([]byte, error) {
+		psCmd := exec.Command("micro", serv.envFlag(), "status")
+		outp, err = psCmd.CombinedOutput()
+		if err != nil {
+			return outp, err
+		}
+
+		if !statusRunning("helloworld", outp) {
```
crufter, line 518: bug here

PR #962, opened 2020-06-22
default namespace changed in go-micro

domwong:


test/test_framework.go
```diff
@@ -15,7 +15,7 @@ import (
 
 const (
 	retryCount = 2
-	isParallel = true
+	isParallel = false
```
crufter, line 18: Why not parallel?

domwong: I'm trying something. I saw locally that tests were flaky if running in parallel. Might be a resource issue on my laptop.

crufter: This flakiness used to be present before with parralel test but haven't seen them recently.

PR #964, opened 2020-06-24
Restore services on restart of runtime

domwong:
closes https://github.com/micro/development/issues/242

### DIFF
--- a/service/runtime/manager/events.go
+++ b/service/runtime/manager/events.go
@@ -19,8 +19,12 @@ var (
 	eventPollFrequency = time.Minute
 )
 
-// eventPrefix is prefixed to the key for event records
-const eventPrefix = "event/"
+const (
+	// eventPrefix is prefixed to the key for event records
+	eventPrefix = "event/"
+	// eventProcessedPrefix is prefixed to the key for tracking event processing
+	eventProcessedPrefix = "processed/"
+)
 
 // publishEvent will write the event to the global store and immediately process the event
 func (m *manager) publishEvent(eType runtime.EventType, srv *runtime.Service, opts *runtime.CreateOptions) error {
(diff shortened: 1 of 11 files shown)


service/runtime/manager/events.go
```diff
@@ -131,6 +131,9 @@ func (m *manager) processEvent(key string) {
 	// write to the store indicating the event has been consumed. We double the ttl to safely know the
 	// event will expire before this record
 	m.cache.Write(&store.Record{Key: key, Expiry: eventTTL * 2})
+	// expire this event from the store
```
ben-toogood, line 134: The store you're writing to here (m.options.Store), is the global KV store which is designed to be consumed by multiple runtimes. Writing to the local cache (as done on L133) should be used to determine if an event has been processed or not.

domwong: hm, OK, didn't realise the relationship here, I'll remove. 

The problem with the local cache is that it doesn't survive restarts since it's an in memory cache which causes events to be replayed in the event of micro server restart so this was to try and get around that. Will need to have a think but will need a way to make this cache persistent either by making it a simple file store or adding some caching file store to reduce read latency.

ben-toogood: The store is the source of truth, i'd record the time at which you read from the store (and should be up to date with all events), then skip processing an event if it was created before this time.

ben-toogood: However, you should probably do more checks against the runtime. e.g. a service could be running in k8s which no longer exists in the store.

domwong: Unfortunately there's a bit of heavy lifting to get a timestamp based version working. We don't even have a way to do reading from stores and ordering by time for instance. For sake of simplicity and MVP I think a simple change to the cache is fine. 

Long term we would rip this out and replace with a proper event streaming solution

ben-toogood: I just also realised, this will error when running in k8s if the service already exists. You might want to perform a Read first and only call Create if you get an ErrNotFound as a result.

test/test_framework.go
```diff
@@ -15,7 +15,7 @@ import (
 
 const (
 	retryCount = 2
-	isParallel = true
+	isParallel = false
```
crufter, line 18: I assume this will be reverted before merge

domwong: indeed. good catch

PR #971, opened 2020-06-26
{cli,proxy} update router support

ben-toogood:
Update the proxy to use the router configure in go-micro (this wasn't previously available). We now set Micro-Network header so the proxy can route to the correct networks.


service/runtime/manager/events.go
```diff
@@ -154,9 +154,9 @@ func (m *manager) runtimeEnv(options *runtime.CreateOptions) []string {
 	// override with vars from the Profile
 	setEnv(m.options.Profile, env)
 
-	// temp: set the auth namespace. this will be removed once he namespace can be determined from certs.
+	// temp: set the service namespace. this will be removed once he namespace can be determined from certs.
 	if len(options.Namespace) > 0 {
-		env["MICRO_AUTH_NAMESPACE"] = options.Namespace
+		env["MICRO_NAMESPACE"] = options.Namespace
```
ben-toogood, line 159: This ENV VAR was renamed in go-micro.

PR #972, opened 2020-06-26
web: implement not found page

ben-toogood:
Currently, micro web returns a 5XX error when a URL cannot be found. This PR changes that to render a 404 page.


client/web/web.go
```diff
@@ -515,6 +512,7 @@ func Run(ctx *cli.Context, srvOpts ...micro.Option) {
 
 	// the web handler itself
 	s.HandleFunc("/favicon.ico", faviconHandler)
+	s.HandleFunc("/notfound", s.notFoundHandler)
```
asim, line 515: /404?

PR #974, opened 2020-06-29
web: update internal resolvers and fix multi-tenancy bugs

ben-toogood:


ben-toogood:
Depends on: https://github.com/micro/go-micro/pull/1756

internal/resolver/web/resolver.go
```diff
@@ -2,150 +2,40 @@ package web
...
 	"strings"
+	"time"
 
+	"github.com/micro/go-micro/v2/api/resolver"
 	res "github.com/micro/go-micro/v2/api/resolver"
-	"github.com/micro/go-micro/v2/client/selector"
 	"github.com/micro/go-micro/v2/registry"
-	"golang.org/x/net/publicsuffix"
+	"github.com/micro/go-micro/v2/router"
+	// "github.com/micro/go-micro/v2/selector"
```
domwong, line 15:
```suggestion
```

internal/resolver/web/resolver.go
```diff
@@ -155,24 +45,35 @@ func (r *Resolver) resolveWithPath(req *http.Request) (*res.Endpoint, error) {
...
-	if err != nil {
-		return nil, err
-	}
+	// select a random route to use
+	// todo: update to use selector once go-micro has updated the interface
+	// route, err := r.Selector.Select(routes...)
+	// if err != nil {
+	// 	return nil, err
+	// }
+	rand.Seed(time.Now().UnixNano())
```
domwong, line 69: Probably don't want to be seeding the global `rand` on every call

ben-toogood: You're right, maybe I should add this to config/cmd?

PR #985, opened 2020-06-30
Added m3o signup test

crufter:
Caveats:

- It's using my own personal stripe test token, sendgrid test token, stripe plan id, stripe template token etc. I have put these into secrets.


go.mod
```diff
@@ -36,3 +38,5 @@ require (
 )
 
 replace google.golang.org/grpc => google.golang.org/grpc v1.26.0
+
+//replace github.com/micro/go-micro/v2 => /home/crufter/code/src/github.com/micro/go-micro
```
ben-toogood, line 39:
```suggestion
```

crufter: :+1:

PR #989, opened 2020-07-01
internal/handler: remove reference to client/selector in rpc test

ben-toogood:


ben-toogood:
Depends on: https://github.com/micro/go-micro/pull/1767

PR #994, opened 2020-07-01
internal/command/cli: scope registry calls to current environment

ben-toogood:
Previously calls such as "micro list services" were listing from all domains, now it's scoped to the current one.


ben-toogood:
Replaced by https://github.com/micro/micro/pull/1000

PR #996, opened 2020-07-01
Support refresh tokens, support otp and basic auth login

crufter:
So far we only had access tokens saved with the CLI, unfortunately this means in 15 minutes one must call `micro call go.micro.auth Auth.Token '{"id":"default","secret":"password"}'` or `micro login email` (in case of m3o) again to acquire an access token.

This PR makes sure that either vanilla `micro server` or the m3o signup flow tokens are properly persisted and used. It also makes `micro login` behave the following way by default:

```
$ micro login
Please enter your email address: email@gmal.com
Enter Password: 
You have been logged in

```

And the m3o signup and login flow can be now accessed under `micro login --otp`.


internal/client/client.go
```diff
@@ -18,15 +26,23 @@ import (
 func New(ctx *ccli.Context) client.Client {
 	env := cliutil.GetEnv(ctx)
 	ns, _ := namespace.Get(env.Name)
-	token, _ := config.Get("micro", "auth", env.Name, "token")
-	return &wrapper{grpc.NewClient(), token, ns, env.ProxyAddress}
+	client := &wrapper{grpc.NewClient(), "", ns, env.ProxyAddress, env.Name, ctx}
```
domwong, line 29: Not a fan of this kind of initialisation, much prefer naming the fields that you're initialising. Makes it very difficult to understand exactly what you're initialising and if the order of params gets changed or a param gets replaced with something else then you could run in to unintended initialisation

crufter: @domudall I'm not either this was already there and usually I try to make smallest deltas possible but will change this.

PR #1003, opened 2020-07-02
service/broker: authorise multi-tenant requests

ben-toogood:


ben-toogood:
Implemented better in a later PR.

PR #1011, opened 2020-07-03
service/registry: de-authenticate the default namespace

ben-toogood:


ben-toogood:
Replaced by a newer PR.

PR #1021, opened 2020-07-06
client/api: bugfix for namespace with versioning

ben-toogood:
Fixes `micro --namespace=foo.api.v1 api` not resolving `/bar` to `foo.api.v1.bar`.

### DIFF
--- a/client/api/api.go
+++ b/client/api/api.go
@@ -195,7 +195,7 @@ func Run(ctx *cli.Context, srvOpts ...micro.Option) {
 
 	// resolver options
 	ropts := []resolver.Option{
-		resolver.WithServicePrefix(Namespace + "." + Type),
+		resolver.WithServicePrefix(apiNamespace),
 		resolver.WithHandler(Handler),
 	}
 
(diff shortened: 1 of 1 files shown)


client/api/api.go
```diff
@@ -85,8 +85,14 @@ func Run(ctx *cli.Context, srvOpts ...micro.Option) {
 		Namespace = strings.TrimSuffix(ctx.String("namespace"), "."+Type)
 	}
 
-	// apiNamespace has the format: "go.micro.api"
-	apiNamespace := Namespace + "." + Type
+	// if the namespace was foo.api.v1, it would excape the trim suffix check
+	// above and we want to use this as our API namespace
+	var apiNamespace string
+	if strings.Contains(Namespace, ".api.") {
```
domwong, line 91: Fine as long as we don't have a domain called "api". Have logged a ticket to make sure we close this off https://github.com/micro/development/issues/270 some time in the future.

ben-toogood: This is service prefix, not the multi-tenant domain, e.g "go.micro"

PR #1024, opened 2020-07-08
Micro namespace no longer special, testing the multitenancy isolation of config values + minor account generation and namespace changes

crufter:

### DIFF
--- a/service/auth/handler/auth/accounts.go
+++ b/service/auth/handler/auth/accounts.go
@@ -14,11 +14,28 @@ import (
 
 // List returns all auth accounts
 func (a *Auth) List(ctx context.Context, req *pb.ListAccountsRequest, rsp *pb.ListAccountsResponse) error {
+	// set defaults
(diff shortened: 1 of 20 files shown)


service/auth/handler/auth/accounts.go
```diff
@@ -22,6 +22,9 @@ func (a *Auth) List(ctx context.Context, req *pb.ListAccountsRequest, rsp *pb.Li
 		req.Options.Namespace = namespace.DefaultNamespace
 	}
 
+	// setup the defaults incase none exist
+	a.setupDefaultAccount(req.Options.Namespace)
```
crufter, line 26: This is rather inelegant but required: default account generation only succeeded in the "micro" namespace but not in custom namespaces. This is an unfortunately consequence of "micro" namespace being so special in many ways.

crufter:
Closing due to staleness and irrelevancy

PR #1030, opened 2020-07-09
Push cell tags, change default image

crufter:
For micro run support

***The push to docker fails currently, needs to be fixed***

### DIFF
--- a/service/runtime/service.go
+++ b/service/runtime/service.go
@@ -48,7 +48,7 @@ var (
 	// DefaultRetries which should be attempted when starting a service
 	DefaultRetries = 3
 	// Image to specify if none is specified
-	Image = "docker.pkg.github.com/micro/cells:go"
+	Image = "micro/cells:go"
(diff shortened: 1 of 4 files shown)


crufter:
Ok the cell wont work as is as it expects the following format:

```
docker run micro/cells:go https://github.com/micro/examples helloworld
```

service/runtime/service.go
```diff
@@ -48,7 +48,7 @@ var (
 	// DefaultRetries which should be attempted when starting a service
 	DefaultRetries = 3
 	// Image to specify if none is specified
-	Image = "docker.pkg.github.com/micro/services"
+	Image = "docker.pkg.github.com/micro/cells:go"
```
asim, line 51: if yo can use the image `micro/cells` it will not require any sort of docker login so will work out of the box for all OSS.

crufter: Niiice

PR #1031, opened 2020-07-09
make m3o tests configurable with env var for src

domwong:
Allows us to inject a different (local) path which allows us to run these tests against different branches of micro/services

### DIFF
--- a/test/m3o_test.go
+++ b/test/m3o_test.go
@@ -47,17 +47,18 @@ func testM3oSignupFlow(t *t) {
 			t.Fatal(string(outp))
 		}
 	}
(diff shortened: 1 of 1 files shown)


test/m3o_test.go
```diff
@@ -195,3 +196,10 @@ func testM3oSignupFlow(t *t) {
 	}
 	wg.Wait()
 }
+
+func getSrcString(envvar, dflt string) string {
```
crufter, line 200: Extreme nit but I think the extra 3 chars defo worth the readability : P

crufter: PS: my bad, default is a keyword : P

PR #1040, opened 2020-07-10
auth: secure the generation of accounts

ben-toogood:


service/auth/handler/auth/auth.go
```diff
@@ -228,6 +244,22 @@ func (a *Auth) Token(ctx context.Context, req *pb.TokenRequest, rsp *pb.TokenRes
 		return errors.BadRequest("go.micro.auth", "Credentials or a refresh token required")
 	}
 
+	// check to see if the secret is a JWT. this is a workaround to allow accounts issued
+	// by the runtime to be refreshed whilst keeping the private key in the server.
+	if a.TokenProvider.String() == "jwt" {
```
crufter, line 249: Now this is the kind of if that I don't like to see :DDD. But since it says "workaround" it's ok :troll: PS: jk

PR #1044, opened 2020-07-12
WIP: V3 experiment - Attempting to move config/cmd and service from go-micro

asim:
First crack at moving go-micro/config/cmd to micro/micro/cmd and moving top level service intialiser to micro/micro/service as a struct which honours the go-micro/service package interface. This might not end up being the final PR but its an attempt at the move to demonstrate where we're going. 

Related PR to strip these things from go-micro https://github.com/micro/go-micro/pull/1821

As I attempt to do this, internally it feels like maybe clients/services should make use of go-micro/service and things we run get templated to use micro/micro/service. Otherwise we might end up in another horrible intialisation problem.


asim:
Closing, taking another shot at it.

PR #1045, opened 2020-07-13
Remove micro init which we no longer use

asim:
We were using `micro init` as a way of updating all the micro server related services running on k8s but likely want to either embed this in the server or move to a different model for self updates.


internal/platform/platform.go
```diff
@@ -1,29 +1,7 @@
...
-	"github.com/micro/micro/v2/internal/update"
-	_ "github.com/micro/micro/v2/internal/usage"
-)
-
 var (
-	// Date of the build
-	// TODO: move elsewhere
-	Version string
-
 	// list of services managed
```
ben-toogood, line 5: Why keep this?

asim: It's used elsewhere. I will refactor it out in a separate PR

PR #1068, opened 2020-07-15
[WIP] Create a platform/cli dir and add a micro signup command

asim:
- Create a platform/cli package for anything plaform specific
- Move the Signup function to platform/cli so `micro login` function only does platform.Signup(ctx) when --otp is set
- Create a `micro signup` command that does exactly what `micro login --otp` does verbatim.

### DIFF
--- a/service/auth/auth.go
+++ b/service/auth/auth.go
@@ -177,12 +177,16 @@ func login(ctx *cli.Context) {
 	}
 
 	authSrv := authFromContext(ctx)
-	password, err := getPassword()
-	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+	password := ctx.String("password")
+	if len(password) == 0 {
+		pw, err := getPassword()
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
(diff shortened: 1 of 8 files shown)


crufter:
Looking good so far. Gotta modify the `m3o_test.go` too as it uses `micro login --otp`.

asim:
We'll also need to ask for password during sign-up and pass to sign-up completion if we want to do it the right way. At that point I think login --otp would probably go away or go via login service...

service/auth/auth.go
```diff
@@ -314,8 +318,8 @@ func Commands(srvOpts ...micro.Option) []*cli.Command {
 					Usage: "Login/signup with a One Time Password.",
 				},
 				&cli.StringFlag{
-					Name:  "email",
-					Usage: "Email address to use for Login with OTP",
+					Name:  "password",
+					Usage: "Password for automated logins",
```
asim, line 322: "Plain text password"

asim:
Closing in favour of https://github.com/micro/micro/pull/1071

PR #1073, opened 2020-07-16
proxy: resolve namespace using token issuer where possible

ben-toogood:


ben-toogood:
Depends on https://github.com/micro/go-micro/pull/1846

PR #1075, opened 2020-07-16
cli: use AcceptsArgs option

ben-toogood:


client/cli/cli.go
```diff
@@ -413,36 +413,6 @@ func StoreCommands() []*cli.Command {
 				},
 			},
 		},
-		{
```
ben-toogood, line 416: These commands are auto-generated now.

cmd/dynamic.go
```diff
@@ -22,10 +22,6 @@ import (
 // the error will also be nil. An error is only returned if there was an issue
 // listing from the registry.
 func lookupService(ctx *cli.Context) (*registry.Service, error) {
```
ben-toogood, line 24: This refactor allowed calling the server also, e.g. "go.micro.store" as well as "go.micro.service.store"

PR #1079, opened 2020-07-16
micro service {auth, router, registry...} refactor

ben-toogood:


client/cli/cli.go
```diff
@@ -480,11 +404,6 @@ func StoreCommands() []*cli.Command {
 //Commands for micro calling action
 func Commands() []*cli.Command {
 	commands := []*cli.Command{
-		{
-			Name:   "cli",
```
asim, line 484: I would leave this here because `micro cli` is a cleaner experience to launch an interactive shell, much like `micro login` or `micro run` are part of a dev UX thing.

asim:
Ok had time to think

1. Let's leave micro {api, web, cli, bot, proxy} where they are from a command perspective
2. Let's move all other services under micro server {broker, config, registry, store, ...}
3. Let's keep the code for services under micro/service/{broker, config, registry, ...} 
4. Let's keep Run, Commands, etc in the per service dirs
5. Let's make sure to modify micro server to run `micro service ...` for the appropriate things

This is probably a good start for the refactor. Then we merge and move on to porting the go-micro/service.go with `type Service struct` implementing the go-micro/service interface and `service.New` returning that.

client/cli/cli.go
```diff
@@ -171,6 +171,82 @@ func NetworkCommands() []*cli.Command {
 	}
 }
 
+//NetworkDNSCommands for networking routing
+func NetworkDNSCommands() []*cli.Command {
```
asim: You can delete these commands. We don't need network dns anymore

cmd/cmd.go
```diff
@@ -351,26 +348,24 @@ func Setup(app *ccli.App, options ...micro.Option) {
 		app.Commands = append(app.Commands, command)
 	}
 
-	// Add the various commands
+	// Add the client commmands
+	app.Commands = append(app.Commands, api.Commands()...)
```
asim, line 352: One thought here which I failed at. I added `Commands` as a variable to this package so that it could be appended to. So each thing could potentially just be imported in main.go with an underscore and let it self register using an init func. One to try.

ben-toogood: Planning to do this in another PR! Completely agree that the modules should self-register commands. I think it's the fact we pass micro options which prevents this.

PR #1084, opened 2020-07-20
cmd: allow any namespace to be used for dynamic commands

ben-toogood:
Previously, only services named "go.micro.service.X" would work with dynamic commands, now it matches using suffix only which means it'll match any, e.g. "micro foo" would map to "go.micro.service.foo", "go.micro.foo", or "bar.service.foo".


cmd/dynamic.go
```diff
@@ -34,20 +30,37 @@ func lookupService(ctx *cli.Context) (*registry.Service, error) {
...
+		return nil, err
+	}
+
+	// filter to services with the correct suffix
+	for _, s := range srvs {
+		if strings.HasSuffix(s.Name, "."+ctx.Args().First()) {
+			srvs, err = reg.GetService(s.Name, registry.GetDomain(dom))
+			if err == nil && len(srvs) > 0 {
+				return srvs[0], nil
+			}
```
domwong, line 63: Repetition here, pull this out in to a func that you can use here and above on line 39. Will also mean it's testable.

PR #1085, opened 2020-07-20
runtime: update usage of runtime credentials option

ben-toogood:
Update to support new secrets option: https://github.com/micro/go-micro/pull/1861


asim:
LGTM

PR #1088, opened 2020-07-20
Kind tests

domwong:


platform/kubernetes/uninstall.sh
```diff
@@ -13,6 +13,16 @@ bash uninstall.sh;
 cd ../nats;
 bash uninstall.sh;
 
+# delete the PVs and PVCs
+for pvc in $(kubectl get pvc -o name | sed 's/persistentvolumeclaim\///g')
+do 
+    kubectl delete pvc $pvc
```
ben-toogood, line 19: you could just do `kubectl delete pvc --all`

domwong: well that's much smarter. must learn to read options.

PR #1113, opened 2020-07-23
Build improvements

domwong:
- Use multistage docker file to reduce image size 977MB -> 511MB 
- Add caching of go/pkg/mod for integration tests
- Use a local registry for Kind to remove dependency on docker hub
- Make `t.Fatal()` return from the test immediately rather than continuing


test/test_framework.go
```diff
@@ -195,7 +197,7 @@ func newLocalServer(t *t, fname string, opts ...options) testServer {
 func (s *testServerBase) launch() error {
 	go func() {
 		if err := s.cmd.Start(); err != nil {
-			s.t.t.Fatal(err)
+			s.t.Fatal(err)
```
crufter, line 200: This will prevent retries though, probably should add a comment here

domwong: as discussed, this should respect the retries

PR #1115, opened 2020-07-23
v3-alpha

ben-toogood:


client/cli/util/util.go
```diff
@@ -7,31 +7,22 @@ import (
...
 )
 
 const (
-	// EnvLocal is a builtin environment, it means services launched
-	// with `micro run` will use default, zero dependency implementations for
-	// interfaces, like mdns for registry.
+	// EnvLocal is a builtin environment, it represents your local `micro server`
 	EnvLocal = "local"
-	// EnvServer is a builtin environment, it represents your local `micro server`
-	EnvServer = "server"
```
domwong, line 24: YES!!!

client/web/web.go
```diff
@@ -435,7 +442,9 @@ func Run(ctx *cli.Context) error {
 	// Setup the web resolver
 	var resolver res.Resolver
 	resolver = &web.Resolver{
-		Router: s.Options().Router,
+		Router: regRouter.NewRouter(
+			router.Registry(muregistry.DefaultRegistry),
```
domwong, line 446: I see a lot of these replacing `Options().Something` with `DefaultSomething`. 

What's the practical impact of this? Does that mean things are less configurable? Not necessarily a bad thing...

ben-toogood: Previously everything was an option on service, which was configured in go-micro. This didn't scale since the service options became huge. Now each package in micro/service has its own default. In the next PR i'll add public accessors, e.g. store.Read() would call the Read method on the DefaultStore. 

If you see the profiles, all they do now is set the defaults across micro in bulk, e.g. setting MICRO_PROFILE=ci will do `registry.DefaultRegistry = etcd.NewRegistry()`. If someone wants to override a specific default, they can add a plugins.go file which can set defaults in the same way.

cmd/util.go
```diff
@@ -0,0 +1,129 @@
...
+	a.Init(
+		auth.ClientToken(token),
+		auth.Credentials(accID, accSecret),
+	)
+	return nil
+}
+
+// refreshAuthTokenPeriodically if it is close to expiring
+func refreshAuthTokenPeriodically() {
+	timer := time.NewTicker(time.Second * 15)
```
domwong, line 105: We should `Stop()` the ticker because it will cause a memory leak if we don't

ben-toogood: Understood, where should we stop it? It lives for the lifetime of the process.

main.go
```diff
@@ -1,29 +1,37 @@
 package main
 
+//go:generate ./.github/generate.sh
```
domwong, line 3: Feel like we should find a better home for this script than `.github`

profile/profile.go
```diff
@@ -0,0 +1,106 @@
+// Package profile is for specific profiles
+// @todo this package is the definition of cruft and
+// should be rewritten in a more elegant way
+package profile
```
domwong, line 4: I think we're painting ourselves in to a corner somewhat with the default globals. Might need some dependency injection solution at some point

server/server.go
```diff
@@ -73,10 +73,16 @@ func init() {
...
+				EnvVars: []string{"MICRO_AUTO_UPDATE"},
+			},
 			&cli.StringFlag{
-				Name:    "profile",
-				Usage:   "Set the runtime profile to use for services e.g local, kubernetes, platform",
-				EnvVars: []string{"MICRO_RUNTIME_PROFILE"},
+				Name:    "update_url",
+				Usage:   "Set the url to retrieve system updates from",
+				EnvVars: []string{"MICRO_UPDATE_URL"},
+				Value:   "https://micro.mu/update",
```
domwong, line 85: I thought we had removed the auto update stuff? 

Is `micro.mu` the right place to check? Or should it be `m3o.com`?

server/server.go
```diff
@@ -118,17 +124,11 @@ func Run(context *cli.Context) error {
...
-	if len(profile) == 0 {
-		profile = "server"
-	}
-
-	env = append(env, "MICRO_RUNTIME_PROFILE="+profile)
-	env = append(env, os.Environ()...)
+	// pass the env to the services, we unset the profile
+	// env var so it doesn't get passed to the clients
+	// e.g. api, web, proxy
+	os.Unsetenv("MICRO_PROFILE")
```
domwong, line 130: Is there a more elegant way we can do this rather than unsetting for everything? Means that we have an implicit ordering now where anything that depends on `MICRO_PROFILE` needs to run before this. code.

server/server.go
asim: https://micro.mu/update has latest commit and latest docker update, micro.mu remaining the OSS url but I think we should remove auto update related stuff until we have a better understanding of how we want it to operate longer term which I believe is centrally around specific environments.

main.go
asim: Moving to scripts dir

server/server.go
asim: Removed. We should revisit when we want to build in auto update of micro and the services on top. The idea is an android style update model where services are updated whenever needed and the server could provide a notification somewhere to ask the user to issue an update with Server.Restart

server/server.go
asim: It sounds like the profile is actually for the server and should be MICRO_SERVER_PROFILE and we should separately apply a service profile.

asim: When I run the server I expect to run it locally or on kubernetes so micro server --profile={local, kubernetes} with variable MICRO_SERVER_PROFILE. When the server runs services I expect to be able to setup those services using a separate value like MICRO_SERVICE_PROFILE. Its just simply a way to aggregate setup values.

profile/profile.go
asim: If we do dependency injection we need to learn from wire https://github.com/google/wire

cmd/util.go
asim: Resolved by adding an After func to the command which closes an exit channel. The refresh func takes this exit channel and stops the timer when the function finishes.

server/server.go
asim: Below its reset for anything that depends on it. We now check for the clients {api, proxy, web, cli, bot} and set MICRO_PROFILE=client

service/config/client/client.go
```diff
@@ -0,0 +1,106 @@
...
+	"net/http"
+
+	"github.com/micro/go-micro/v3/config/source"
+	"github.com/micro/go-micro/v3/errors"
+	"github.com/micro/go-micro/v3/logger"
+	"github.com/micro/micro/v3/service/client"
+	proto "github.com/micro/micro/v3/service/config/proto"
+)
+
+var (
```
domwong, line 14: const?

service/store/cli/util.go
```diff
@@ -56,9 +55,10 @@ func makeStores(ctx *cli.Context) (store.Store, store.Store, error) {
...
-	if !exists {
-		return nil, errors.Errorf("store %s is not an implemented store - check your plugins", s)
-	}
-	return builtinStore, nil
+	// builtinStore, exists := cmd.DefaultStores[s]
+	// if !exists {
+	// 	return nil, errors.Errorf("store %s is not an implemented store - check your plugins", s)
+	// }
+	// return builtinStore, nil
+	return nil, nil
```
domwong, line 63: eh? Why don't we just nuke this?

test/Dockerfile
```diff
@@ -24,15 +24,9 @@ ADD         scripts/run-etcd.sh /bin/run.sh
...
-COPY ./test/config-example-service ${GOPATH}/src/config-example-service
-
 # Speeding up tests by predownloading dependencies for services used.
-RUN cd ${GOPATH}/src/example-service && go get && \
-    cd ${GOPATH}/src/config-example-service && go get && \
-    go get github.com/micro/examples/helloworld
-
-
+RUN go get github.com/micro/services
+COPY . .
```
domwong, line 29: Why are we copying and building everything here? We were previously building outside of the Dockerfile which made it faster for for testing locally.

test/runtime_test.go
```diff
@@ -182,80 +180,6 @@ func testRunAndKill(t *t) {
 	}
 }
 
-func TestLocalOutsideRepo(t *testing.T) {
```
domwong, line 185: Why was this test removed?

PR #1120, opened 2020-07-24
Make t.Fatal goroutine safe

domwong:
the stdlib `t.Fatal()` doesn't exit immediately when called in goroutines. Fix up our impl to behave similarly (and fix an issue in the process)


crufter:
Merged this to https://github.com/micro/micro/pull/1118

PR #1125, opened 2020-07-27
[WIP] Make uploading local files to micro server work on k8s

crufter:
- [x] Files saved are now persisted in the store, local files are only used to speed up upload/download 
  unlike files the store does not support write and read at certain offset efficiently
- [x] Added namespace support to file uploads/downloads
- [x] Made default runtime download the file from the server
- [ ] Made k8s/cells download file from the server
- [ ] Made client connect to the same node (ie. sticky sessions)


crufter:
We decided to take a different route, closing.

PR #1133, opened 2020-07-29
Do not use service token by default

crufter:
Main part is here https://github.com/micro/micro/pull/1133/files#diff-72b64e133c1e2c1a2627c7b68999e171R43


go.mod
```diff
@@ -21,6 +19,7 @@ require (
 	github.com/micro/cli/v2 v2.1.2
 	github.com/micro/go-micro/v2 v2.9.1
 	github.com/micro/go-micro/v3 v3.0.0-alpha.0.20200728125458-9813f98c8b60
+	github.com/micro/micro v1.18.0 // indirect
```
asim, line 22: micro v1 import

crufter: This keeps creeping back I think it's due to VScode. I usually do a go mod tidy before finalizing the PR, that makes it go away.

service/registry/client/client.go
```diff
@@ -23,6 +23,7 @@ type srv struct {
 
 func (s *srv) callOpts() []client.CallOption {
 	var opts []client.CallOption
+	opts = append(opts, client.WithServiceToken())
```
crufter, line 26: We gotta make sure these options are only can be set in the server side.

internal/wrapper/wrapper.go
```diff
@@ -44,6 +44,9 @@ func (a *authWrapper) wrapContext(ctx context.Context, opts ...client.CallOption
 	if _, ok := metadata.Get(ctx, "Authorization"); ok && !options.ServiceToken {
```
crufter, line 44: Should just delete this if entirely

PR #1155, opened 2020-07-31
Multi Config + server.Command

asim:
- Adds support for specifying config file via -c
- Adds Command() to test/Server so that its a pre-initialised server with its own independent config and env flags.


domwong:
Merged and will fix the final integration test in a separate PR

PR #1204, opened 2020-08-07
WIP: fix for router init chan

asim:
This PR fixes an initChan bug in the go-micro registry router. initChan was defined in a select where default is actually blocking. This has been moved out but has introduced new problems which were hard to diagnose.

At the moment the below error is being managed by switching the log level of this statement to warn so its not output. This is being return in the CLI highlighting that the CLI should be a purely rpc thing and have a different setup path to everything else. We can also see some side effective from existing defaults being initialised in go-micro options.

```    
test_framework.go:158: store_test.go:29, Calling micro store read (failed after 2m0.504244793s with 'store read should fail'), output: '2020-08-08 08:40:30  file=registry/registry.go:590 level=warn Error watching the registry: rpc error: code = PermissionDenied desc = {"Id":"registry.Registry.Watch","Code":403,"Detail":"Access denied to namespace","Status":"Forbidden"}
```

TODO: separate out client initialisation and avoid having anything setup before its explicitly defined.


asim:
Close in favour of https://github.com/micro/micro/pull/1208/files

PR #1215, opened 2020-08-10
{cmd/service}: add service_address flag

ben-toogood:
Add a service_address flag, which allows setting the port to bind to. This will enable readiness checks in k8s.


ben-toogood:
Extended in https://github.com/micro/micro/pull/1217

PR #1221, opened 2020-08-10
WIP: Plugin design iteration for wrappers and all things

asim:
Micro includes a plugin package which can be used to augment micro itself. Things like wrappers and extra functionality should be modular plugins rather than being baked into cmd/cmd. This keep it clean and allows us to just import _ things as needed. Ideally we move mostly everything out of cmd/cmd and either into main.go or service.go


ben-toogood:
I think this would be a big improvement

asim:
Closing, will leave as a link in a m3o/dev issue for later

PR #1236, opened 2020-08-11
bump go mod

domwong:


domwong:
Closing, go-micro will be bumped as part of a different in flight PR anyway

PR #1238, opened 2020-08-12
test possible k8s test fix

ben-toogood:


test/kind.go
```diff
@@ -81,17 +81,30 @@ func (s *testK8sServer) Run() error {
...
 	}
 
-	ChangeNamespace(s.Command(), s.Env(), s.Env())
-	if s.opts.Login {
-		Login(s, s.t, s.Env(), "micro")
+	// login with this new account
+	Login(s, s.t, s.Env(), "micro")
+
+	// delete the admin account
+	outp, err := s.Command().Exec("auth", "delete", "account", "admin")
```
domwong, line 105: I thought an account is global? e.g. there is only 1 account called `admin` regardless of namespace.

ben-toogood: No, accounts are scoped per namespace.

service/auth/server/auth/auth.go
```diff
@@ -121,7 +121,7 @@ func (a *Auth) Generate(ctx context.Context, req *pb.GenerateRequest, rsp *pb.Ge
 		req.Options = &pb.Options{}
 	}
 	if len(req.Options.Namespace) == 0 {
-		req.Options.Namespace = namespace.DefaultNamespace
+		req.Options.Namespace = namespace.FromContext(ctx)
```
domwong, line 124: What changed such that we needed to make this fix? Why wasn't it picked up in previous tests?

ben-toogood: This isn't needed for this PR but it's more standard with the other handlers.

PR #1240, opened 2020-08-12
Metadata fixes: updated time, owner, group

crufter:


service/runtime/server/handler.go
```diff
@@ -136,6 +142,23 @@ func (r *Runtime) Update(ctx context.Context, req *pb.UpdateRequest, rsp *pb.Upd
...
+		}
+		service.Metadata["owner"] = account.ID
+		// This is a hack - we don't want vanilla `micro server` users where the auth is noop
+		// to have long uuid as owners, so we put micro here - not great, not terrible.
+		if auth.DefaultAuth.String() == "noop" {
+			service.Metadata["owner"] = "micro"
+		}
+		service.Metadata["group"] = account.Issuer
+	}
+	service.Metadata["started"] = time.Now().Format(time.RFC3339)
```
ben-toogood, line 159: service.Metadata could be nil

service/runtime/server/handler.go
```diff
@@ -136,6 +142,23 @@ func (r *Runtime) Update(ctx context.Context, req *pb.UpdateRequest, rsp *pb.Upd
 	return nil
 }
 
+func setupServiceMeta(ctx context.Context, service *runtime.Service) {
+	account, accOk := goauth.AccountFromContext(ctx)
+	if accOk && account != nil {
```
ben-toogood, line 147: Account should never be nil if accOK is true

service/runtime/server/handler.go
crufter: There is a check for that above

crufter: ... at the wrong place

service/runtime/server/handler.go
crufter: Gotcha

PR #1252, opened 2020-08-13
Change password functionality

crufter:
- [x] Change secret endpoint
- [x] Change password CLI command
- [x] Test


client/cli/user/user.go
```diff
@@ -37,11 +44,96 @@ func init() {
...
+			if pw != pwVer {
+				fmt.Println("Passwords do not match. Please try again.")
+				continue
+			}
+			newPassword = pw
+			break
+		}
+	}
+
+	// Inspect the token
```
ben-toogood, line 117: Why do we not do this before asking for passwords?

client/cli/user/user.go
```diff
@@ -37,11 +44,96 @@ func init() {
...
+	}
+
+	accountService := pb.NewAccountsService("auth", client.DefaultClient)
+	_, err = accountService.ChangePassword(context.TODO(), &pb.ChangePasswordRequest{
+		Id:        acc.ID,
+		OldSecret: oldPassword,
+		NewSecret: newPassword,
+	}, goclient.WithAuthToken())
+	if err != nil {
+		fmt.Println(err)
```
ben-toogood, line 131: Rather than printing the error and exiting, you can just return the error.

client/cli/user/user.go
```diff
@@ -37,11 +44,96 @@ func init() {
...
+	// Inspect the token
+	acc, err := auth.Inspect(token)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	accountService := pb.NewAccountsService("auth", client.DefaultClient)
+	_, err = accountService.ChangePassword(context.TODO(), &pb.ChangePasswordRequest{
+		Id:        acc.ID,
```
ben-toogood, line 126: This should probably be determined using the account.FromContext() function in the handler, as apposed to be passed from the client

service/auth/proto/auth.proto
```diff
@@ -147,4 +148,14 @@ message ListRequest {
 
 message ListResponse {
 	repeated Rule rules = 1;
+	string a = 2;
```
ben-toogood, line 151: I think this is a typo?

client/cli/user/user.go
crufter: Good point

service/auth/proto/auth.proto
```diff
@@ -147,4 +148,14 @@ message ListRequest {
...
 }
+
+message ChangePasswordRequest{
```
ben-toogood, line 154: Mixed terminology: password vs secret.

client/cli/user/user.go
crufter: I will look that up thanks.

service/auth/server/auth/accounts.go
```diff
@@ -115,6 +116,55 @@ func (a *Auth) Delete(ctx context.Context, req *pb.DeleteAccountRequest, rsp *pb
...
+// ChangePassword by providing a refresh token and a new secret
+func (a *Auth) ChangePassword(ctx context.Context, req *pb.ChangePasswordRequest, rsp *pb.ChangePasswordResponse) error {
+	// set defaults
+	if req.Options == nil {
+		req.Options = &pb.Options{}
+	}
+	if len(req.Options.Namespace) == 0 {
+		req.Options.Namespace = namespace.DefaultNamespace
+	}
+
```
ben-toogood, line 128: You need to authorise access to the namespace using the namespace.Authorize function. See List for example usage.

service/auth/proto/auth.proto
crufter: yeah when i was trying to get proto working... fixing.

service/auth/proto/auth.proto
crufter: Yeah makes sense.

service/auth/server/auth/accounts.go
crufter: +1

client/cli/user/user.go
crufter: Changed based on our discussion

service/auth/proto/auth.proto
crufter: Changed

service/auth/server/auth/accounts.go
crufter: Done

PR #1255, opened 2020-08-14
[WIP] Snap for goreleaser

crufter:


asim:
Parking until distribution of installation becomes a priority e.g more user requests for snap, yum, apt and brew.

PR #1256, opened 2020-08-14
Baseurl, gitlab support, single word service names

crufter:
- [x] Base urls can now be specified with `micro user config set git.baseurl` or `micro user config set git.[envname].baseurl`
- [x] Service names are now "one word", ie. "test/service/config-example" -> "config-example"
- [x] Gitlab support

https://github.com/micro/go-micro/pull/1933

```
$ micro run test/service/config-example/
$ micro status
NAME		    VERSION	SOURCE					            STATUS	BUILD	UPDATED	METADATA
config-example	latest	config-example.tar.gz/test/service	running	n/a	    11s ago	owner=micro, group=micro
```


test/runtime_test.go
```diff
@@ -588,33 +613,36 @@ func testRunPrivateSource(t *T) {
...
+		if !statusRunning("helloworld", "latest", outp) {
 			return outp, errors.New("Can't find helloworld service in runtime")
 		}
 		return outp, err
-	}, 120*time.Second); err != nil {
+	}, 60*time.Second); err != nil {
 		return
 	}
 
+	outp, err := cmd.Exec("logs", "helloworld")
```
ben-toogood, line 624: Remove debugging

crufter: Good catch

test/runtime_test.go
```diff
@@ -269,6 +257,44 @@ func testRunGithubSource(t *T) {
...
+	cmd := serv.Command()
+	cmd.Exec("user", "config", "set", "git."+serv.Env()+".baseurl", "gitlab.com/micro-test")
+
+	outp, err := cmd.Exec("run", "basic-micro-service")
+	if err != nil {
+		t.Fatalf("micro run failure, output: %v", string(outp))
+		return
+	}
+
+	if err := Try("Find helloworld in runtime", t, func() ([]byte, error) {
```
ben-toogood, line 283: Can you please add an additional test to check that the service registered. If there was an error pulling source it might not be caught by this test.

test/runtime_test.go
```diff
@@ -269,6 +257,44 @@ func testRunGithubSource(t *T) {
...
+	t.Parallel()
+
+	serv := NewServer(t, WithLogin())
+	defer serv.Close()
+	if err := serv.Run(); err != nil {
+		return
+	}
+
+	cmd := serv.Command()
+	cmd.Exec("user", "config", "set", "git."+serv.Env()+".baseurl", "gitlab.com/micro-test")
```
ben-toogood, line 275: Should be a repo, not an org/user

ben-toogood: Also, can you please add this value to the "micro user config" output.

test/runtime_test.go
crufter: Good idea

test/runtime_test.go
crufter: > Should be a repo, not an org/user
It works for both cases, I put the code in the top level to have a test for this too and it works. Might make sense to have a test for subfolder cases too though.

crufter: > Also, can you please add this value to the "micro user config" output.
Nice, I was not familiar with that command

crufter: Done

PR #1261, opened 2020-08-18
Add more tests for store

domwong:
This is the start of adding more tests for store. These tests sit at the micro rather than go-micro or micro CLI level so primarily they are testing that the code/packages, which our users will be coding against, work as expected. 

The way we approach the tests is to create a new service in test/service which has endpoints that exercise whatever code paths we wish to test. Then in our integration tests we simply spin up a micro server, run the service, and then call the endpoint with the endpoint itself actually deciding whether the test has passed or failed. One of the reasons we do this is because the bootstrapping of as service is quite complex and so it's actually easier to just create a new service. The side effect of this is that we can start running these tests on both k8s and local platforms.

This PR also adds a `--request_timeout` arg to `micro call` which is useful for calling the test services.


test/test_framework.go
```diff
@@ -141,8 +141,6 @@ func (c *Command) Output() ([]byte, error) {
 func Try(blockName string, t *T, f cmdFunc, maxTime time.Duration) error {
 	// hack. k8s can be slow locally
 	maxTime *= maxTimeMultiplier
-	// backoff, the retry logic is basically to cover up timing issues
-	maxTime *= time.Duration(t.attempt)
```
domwong, line 145: This was a blunt tool to get us over some issues. Removing and will be replaced with something better soon.

internal/command/command.go
```diff
@@ -234,6 +234,13 @@ func CallService(c *cli.Context, args []string) ([]byte, error) {
 	creq := client.NewRequest(service, endpoint, request, goclient.WithContentType("application/json"))
 
 	opts := []goclient.CallOption{goclient.WithAuthToken()}
+	if timeout := c.String("request_timeout"); timeout != "" {
```
crufter, line 237: Loving this

PR #1262, opened 2020-08-18
Support private repos for env 'local'

crufter:
- [x] Fix swallowing of runtime.Create errors in the runtime manager
- [x] Support github private repos for env local
- [x] Support gitlab private repos for env local
- [x] Made github private test run in normal integration flow
- [x] Added gitlab private test
- [x] Added test for single repo deploys <-- was actually done already*
- [x] Testing updates with private repos

*We should close https://github.com/m3o/dev/issues/382
Go micro issue https://github.com/micro/go-micro/pull/1938


service/runtime/manager/events.go
```diff
@@ -117,7 +118,8 @@ func (m *manager) processEvent(key string) {
 		err = runtime.Update(ev.Service, gorun.UpdateNamespace(ns))
 	case gorun.Create:
 		// generate an auth account for the service to use
-		acc, err := m.generateAccount(ev.Service, ns)
```
crufter, line 120: Error was being created here as a local variable which caused create errors to be swallowed.

PR #1266, opened 2020-08-18
platform/kubernetes: deploy NATS streaming

ben-toogood:


ben-toogood:
Waiting for a response from the NATs team regarding an issue with replicas being fixed at 3.

ben-toogood:
> Waiting for a response from the NATs team regarding an issue with replicas being fixed at 3.

This was a confusion on my part, nats runs three containers per pod however it's not three replicas of the nats server.

crufter (approved):

PR #1276, opened 2020-08-19
WIP DO NOT MERGE: Runtime source Upload/Download endpoints

asim:
This is an example of adding Source.Upload/Download endpoints for the runtime. In this case rather than using the very complex go-micro/util/file as an endpoint on server, we move to runtime actually managing the source to running part (even where it may offload builds on the backend in future). Our goal is to simply stream chunks of a tarball using a grpc stream from local to runtime and again download in the same manner where necessary.

Ideally in the "micro" cell we could call `micro source download path/to/source` which would encapsulate the streaming of a file and unpacking it or `micro run source` in the cell would actually work as a local pull doing the same. Whatever makes sense.

Ignore internal/git the goal was to try pull stuff out of go-micro

https://github.com/m3o/dev/issues/231


asim:
Was for demonstration purposes

PR #1291, opened 2020-08-21
implement runtime delete/create namespace

domwong:


asim (approved):

domwong:
merged as #1298  because I had to rename the branch

PR #1300, opened 2020-08-26
Self update binary if a new release is available

asim:
Prompts the user to update the binary if a new release is available. Only does this on CLI commands, not for "server" or "service" commands. It means the release can stay up to date with the platform.

https://github.com/m3o/dev/issues/397


cmd/update.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/blang/semver"
+	"github.com/rhysd/go-github-selfupdate/selfupdate"
+)
+
+// confirmAndSelfUpdate looks for a new release of micro and upgrades in place
+// we only execute this for select CLI commands rather than everything
+func confirmAndSelfUpdate() (bool, error) {
+	latest, found, err := selfupdate.DetectLatest("micro/micro")
+	if err != nil {
+		return false, fmt.Errorf("Error occurred while detecting version:", err)
```
domwong, line 17:
```suggestion
		return false, fmt.Errorf("Error occurred while detecting version: %s", err)
```

cmd/update.go
```diff
@@ -0,0 +1,46 @@
...
+		return false, err
+	}
+
+	v := semver.MustParse(version)
+	if !found || latest.Version.LTE(v) {
+		// current version is the latest
+		return false, nil
+	}
+
+	fmt.Print("Do you want to update to", latest.Version, "? (yes/no): ")
```
domwong, line 27:
```suggestion
	fmt.Print("Do you want to update to ", latest.Version, "? (yes/no): ")
```

cmd/update.go
```diff
@@ -0,0 +1,46 @@
...
+	if input == "no\n" {
+		return false, nil
+	}
+
+	exe, err := os.Executable()
+	if err != nil {
+		return false, fmt.Errorf("Could not locate executable path")
+	}
+	if err := selfupdate.UpdateTo(latest.AssetURL, exe); err != nil {
+		return false, fmt.Errorf("Error occurred while updating binary:", err)
```
domwong, line 41:
```suggestion
		return false, fmt.Errorf("Error occurred while updating binary: %s", err)
```

domwong (approved):
Haven't tried it, code looks ok

asim:
> Haven't tried it, code looks ok

Hard to test without a new release. Not sure what to do? Also if we don't update the internal version in cmd/version.go it also becomes problematic. What's the best way to stay on top of this?

PR #1321, opened 2020-09-03
Don't delete people delete their own account

crufter:


domwong:
This wouldn't actually have solved Lucas's issue. He created another admin account and deleted his proper one. I think the fix should include not being able to delete yourself

service/auth/server/auth/accounts.go
```diff
@@ -93,6 +93,18 @@ func (a *Auth) Delete(ctx context.Context, req *pb.DeleteAccountRequest, rsp *pb
 		return errors.BadRequest("auth.Accounts.Delete", "Error querying accounts: %v", err)
 	}
 
+	acc, ok := auth.AccountFromContext(ctx)
+	if !ok {
+		return errors.Unauthorized("auth.Accounts.Delete", "Unauthorized")
+	}
+	switch {
+	case acc.Issuer == namespace.DefaultNamespace:
```
crufter, line 101: Not sure if this is needed at all.

domwong (approved):

PR #1324, opened 2020-09-04
Using JWT auth for local. Generate JWT keys on the fly if not provided.

crufter:


internal/user/user.go
```diff
@@ -0,0 +1,156 @@
...
+
+func init() {
+	user, err := user.Current()
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	Dir = filepath.Join(user.HomeDir, path)
+}
+
+func GetKeys() ([]byte, []byte, error) {
```
crufter, line 32: This should not be here as it's server specific. Suggestions welcome

internal/user/user.go
```diff
@@ -0,0 +1,160 @@
...
+	"encoding/base64"
+	"encoding/pem"
+	"io/ioutil"
+	"os"
+
+	"github.com/micro/micro/v3/service/logger"
+)
+
+var (
+	Dir  = ""
```
crufter, line 20: Probably should be called microdir

asim: user.Dir is fine

internal/user/user.go
asim: Don't worry too much for now but can be refactored to internal/server if need be.

asim (approved):

internal/config/config.go
```diff
@@ -157,7 +152,7 @@ func newConfig() (conf.Config, error) {
 
 	// now write the file if it does not exist
 	if _, err := os.Stat(path); os.IsNotExist(err) {
-		ioutil.WriteFile(path, []byte(`{}`), 0644)
+		ioutil.WriteFile(path, []byte(`{"env":"local"}`), 0644)
```
crufter, line 155: Have noticed `micro user token`returns nothing as env is empty after I nuke `~/.micro`. This fixes that.

internal/user/user.go
```diff
@@ -33,7 +33,10 @@ func init() {
 	}
 }
 
-func GetKeys() ([]byte, []byte, error) {
+// GetJWTCerts returns local keys or generates and returns them for JWT auth.GetJWTCerts
+// This is only here for "0 dep", so people don't have to create and load the certs themselves,
+// not really intended for serious production use.
+func GetJWTCerts() ([]byte, []byte, error) {
```
asim, line 39: The previous name was fine. Its a keypair. So GetKeys made sense. The only thing to do would be to move to an internal/server package.

PR #1328, opened 2020-09-07
Client SDK generator

asim:
Something makes me think clients generated from protos should live in this repo. Maybe I'm wrong, I don't know. Realistically clients are published to various locations like npm, ruby gems, etc. So their source location doesn't really matter, but considering the protos are here, it's helpful.


asim:
Leaving totally separate for now

PR #1334, opened 2020-09-08
separate ci/platform profiles and generate using a micro init command

asim:
`micro init` (potentially to be renamed to micro gen) will initialise a profile.go file with profiles that can be added at build time. A previous iteration was to generate a plugins.go file but we're now now moving on to profiles. 

CI and Platform provides are now versioned separately using a different go mod. This means its removed from the normal work flow. Micro can be now be downloaded with go get as it removes the etcd problem. This also means user code is much more lightweight, when they import micro, they aren't importing all our infra deps. Only the plaform profile makes use of this.

We then separately build a micro/platform binary, rather than conflating it all into micro/micro.

Closes https://github.com/m3o/dev/issues/362


ben-toogood:
Related: https://github.com/m3o/dev/issues/362

test/Dockerfile
```diff
@@ -22,12 +22,5 @@ VOLUME      /data
 EXPOSE      2379 2380 4001 7001
 ADD         scripts/run-etcd.sh /bin/run.sh
 
-ENV MICRO_REGISTRY=etcd
-
-# Speeding up tests by predownloading and building dependencies for services used.
-COPY . .
-RUN go get github.com/micro/services
-RUN go get github.com/micro/services/helloworld
-RUN bash -c 'for d in $(find test -name "main.go" | xargs -n 1 dirname); do pushd $d && go get && popd; done'
```
domwong, line 31: This was there to speed up the k8s tests by downloading dependencies in to the image that runs the services

ben-toogood: It now happens in the github workflow

domwong: ignore me. its a separate kind file now

domwong (approved):

PR #1341, opened 2020-09-09
Add username as alias for account ID in auth

domwong:
Allow a user to login either using ID or username as the account identifier. This change allows us to use ID for some stable identifier e.g. UUID and username as a user friendly alias which could change e.g. email etc.

This means that the account ID can be used as a stable reference key in other parts of system where required and the username can be used as a user friendly string which can be changed when required. NOTE the mechanism to change is out of the scope of this PR.


test/auth_test.go
```diff
@@ -253,3 +253,47 @@ func curl(serv Server, path string) (map[string]interface{}, error) {
...
+func testUsernameLogin(t *T) {
+	t.Parallel()
+	serv := NewServer(t, WithLogin())
+	defer serv.Close()
+	if err := serv.Run(); err != nil {
+		return
+	}
+
+	cmd := serv.Command()
+	outp, err := cmd.Exec("call", "auth", "Auth.Generate", `{"id":"someID", "name":"someUsername", "secret":"password"}`)
```
ben-toogood, line 271: You could do `"auth", "generate" "--id=someID"` etc

test/auth_test.go
```diff
@@ -253,3 +253,47 @@ func curl(serv Server, path string) (map[string]interface{}, error) {
...
+	if err != nil {
+		t.Fatalf("Error generating account %s %s", string(outp), err)
+	}
+	outp, err = cmd.Exec("login", "--username", "someUsername", "--password", "password")
+	if err != nil {
+		t.Fatalf("Error logging in with user name %s %s", string(outp), err)
+	}
+	outp, err = cmd.Exec("login", "--username", "someID", "--password", "password")
```
ben-toogood, line 279: Might be worth testing the email alias here also since that's still in the docs

ben-toogood (approved):

PR #1347, opened 2020-09-11
service: remove leaking go-micro references

ben-toogood:
Closes https://github.com/m3o/dev/issues/391


service/auth/auth.go
```diff
@@ -1,44 +1,84 @@
...
+	DefaultAuth auth.Auth = client.NewAuth()
+	// ErrInvalidToken is when the token provided is not valid
+	ErrInvalidToken = auth.ErrInvalidToken
+	// ErrForbidden is when a user does not have the necessary scope to access a resource
+	ErrForbidden = auth.ErrForbidden
+)
+
+type (
+	// AccessToken is an alias for auth.Token
+	AccessToken = auth.Token
```
asim, line 26: Because AccessToken is a field maybe this would look better as AccountToken?

asim (approved):

PR #1366, opened 2020-09-15
[WIP] Increase timeout

crufter:
Noticed tests failing, it might be because I have changed timeouts recently from 50 secs to 20.


crufter:
This has already been changed in master.

PR #1368, opened 2020-09-16
Blob store

ben-toogood:


service/store/client/blob.go
```diff
@@ -0,0 +1,155 @@
...
+			},
+			Blob: buffer,
+		}
+
+		if err := stream.Send(req); err != nil {
+			return err
+		}
+	}
+
+	return stream.Close()
```
domwong, line 117: Do we need to do a stream.Close() in all return paths?

service/store/client/blob.go
```diff
@@ -0,0 +1,155 @@
...
+		} else if err != nil {
+			return err
+		}
+
+		req := &pb.BlobWriteRequest{
+			Key: key,
+			Options: &pb.BlobOptions{
+				Namespace: options.Namespace,
+			},
+			Blob: buffer,
```
domwong, line 109: I think you need to record the number of bytes read in `blob.Read()` and then send that many bytes from the buffer otherwise on the last write you might send more than you intend to

See https://tour.golang.org/methods/21

service/store/server/blob.go
```diff
@@ -0,0 +1,148 @@
...
+	} else if err == gostore.ErrMissingKey {
+		return errors.BadRequest("store.Blob.Read", "Missing key")
+	} else if err != nil {
+		return errors.InternalServerError("store.Blob.Read", err.Error())
+	}
+
+	// read from the blob and stream it to the client
+	buffer := make([]byte, bufferSize)
+	for {
+		if _, err := blob.Read(buffer); err == io.EOF {
```
domwong, line 50: ditto above, you should record how much is read here and send that many bytes below

service/store/server/blob.go
```diff
@@ -0,0 +1,148 @@
...
+		return errors.BadRequest("store.Blob.Write", "No blob was sent")
+	}
+
+	// parse the options
+	if options == nil || len(options.Namespace) == 0 {
+		options = &pb.BlobOptions{Namespace: namespace.FromContext(ctx)}
+	}
+
+	// authorize the request
+	if err := authns.Authorize(ctx, options.Namespace); err == authns.ErrForbidden {
```
domwong, line 102: anyway we can do this authorization before receiving the data?

ben-toogood: I couldn't find any way, because we pass the namespace etc in the options and that is part of the request payload. I could rely on passing it via the req headers but that feels hacky.

domwong: Thought as much. I wonder if we have to pull it up in to the loop where you're assigning to `options` . From an end user's POV it will give the best experience (e.g. fail fast). Admittedly looks a bit fugly from a code perspective

crufter (approved):

PR #1371, opened 2020-09-17
Config rework

crufter:


go.mod
```diff
@@ -5,23 +5,20 @@ go 1.13
...
 	github.com/go-acme/lego/v3 v3.4.0
+	github.com/gobwas/pool v0.2.1 // indirect
 	github.com/golang/protobuf v1.4.2
 	github.com/google/uuid v1.1.2
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
 	github.com/hashicorp/go-version v1.2.1
 	github.com/juju/fslock v0.0.0-20160525022230-4d5c94c67b4b
-	github.com/micro/go-micro/v3 v3.0.0-beta.2.0.20200918132042-8975184b88a7
+	github.com/micro/go-micro v1.18.0
```
domwong, line 19: Why has a v1 import crept in ? need to track this down and fix to v3

crufter: It's VScodde

service/config/client/options.go
```diff
@@ -3,7 +3,7 @@ package client
 import (
 	"context"
 
-	"github.com/micro/go-micro/v3/config/source"
+	"github.com/micro/go-micro/config/source"
```
domwong: should be v3

service/config/client/client.go
```diff
@@ -14,92 +15,67 @@ import (
...
-		return
+		logger.Error(err)
 	}
-	return newWatcher(stream)
 }
 
-// Write is unsupported
-func (m *srv) Write(cs *source.ChangeSet) error {
-	return nil
+func (m *srv) Delete(path string, options ...config.Option) {
```
domwong, line 56: Think I'm somewhat late to the party here but why don't we return `error` in any of these methods? Just logging error will likely get lost

service/config/client/options.go
crufter: Just fixed, this was what caused 1.18, nice catch

service/config/client/client.go
crufter: I don't know honestly, I kept it as this is how it worked forever but if we change it we should change it now probably.

service/config/server/encryption.go
```diff
@@ -0,0 +1,70 @@
...
+
+// encrypt/decrypt functions are taken from https://www.melvinvivas.com/how-to-encrypt-and-decrypt-data-using-aes/
+
+func encrypt(stringToEncrypt string, key []byte) (encryptedString string) {
+	plaintext := []byte(stringToEncrypt)
+
+	//Create a new Cipher Block from the key
+	block, err := aes.NewCipher(key)
+	if err != nil {
+		panic(err.Error())
```
domwong, line 20: why panic rather than bubble up an error?

crufter: copypasta code that's why :trollface:

service/config/server/handler.go
```diff
@@ -23,432 +21,220 @@ const (
...
-			return errors.BadRequest("config.Config.Update", "unmarshal value error: %v", err)
-		}
-		record = records[0]
-	}
-
-	// generate a new base changeset
-	changeSet := &source.ChangeSet{
-		Format: "json",
-		Data:   []byte(`{}`),
+	ch, err := store.Read(req.Namespace)
```
domwong, line 173: should be `ns` rather than `req.Namespace`

service/config/server/handler.go
```diff
@@ -23,432 +21,220 @@ const (
...
-
-	if err := store.Write(record); err != nil {
-		return errors.BadRequest("config.Config.Update", "update into db error: %v", err)
 	}
 
-	_ = publish(ctx, &pb.WatchResponse{Namespace: req.Change.Namespace, ChangeSet: req.Change.ChangeSet})
-
-	return nil
+	return store.Write(&store.Record{
+		Key:   req.Namespace,
```
domwong, line 206: should be `ns`

service/config/server/handler.go
```diff
@@ -23,432 +21,220 @@ const (
...
-	if len(req.Namespace) == 0 {
-		req.Namespace = defaultNamespace
-	}
-
-	// authorize the request
-	if err := namespace.Authorize(ctx, req.Namespace); err == namespace.ErrForbidden {
-		return errors.Forbidden("config.Config.Watch", err.Error())
-	} else if err == namespace.ErrUnauthorized {
-		return errors.Unauthorized("config.Config.Watch", err.Error())
+	ch, err := store.Read(req.Namespace)
```
domwong, line 226: should be ns

service/config/server/handler.go
```diff
@@ -23,432 +21,220 @@ const (
...
-	return reader.Values(ch)
-}
-
-// publish a change
-func publish(ctx context.Context, ch *pb.WatchResponse) error {
-	req := muclient.NewMessage(watchTopic, ch)
-	return muclient.Publish(ctx, req)
+	values.Delete(req.Path)
+	return store.Write(&store.Record{
+		Key:   req.Namespace,
```
domwong, line 237: should be `ns`

domwong (approved):

PR #1391, opened 2020-09-24
Passing on secret option in config client

crufter:


service/config/client/client.go
```diff
@@ -24,10 +24,15 @@ type srv struct {
...
 func (m *srv) Get(path string, options ...config.Option) (config.Value, error) {
+	o := config.Options{}
+	for _, option := range options {
+		option(&o)
+	}
 	nullValue := config.NewJSONValue([]byte("null"))
 	req, err := m.client.Get(context.DefaultContext, &proto.GetRequest{
 		Namespace: m.namespace,
 		Path:      path,
+		Secret:    o.Secret,
```
asim, line 35: If its an option it should be in options

crufter: I will change config endpoint interface then, but tests need this fix as they fail for Ben now.

crufter: BTW, I think we should consider adopting the convention of top level fields for endpoints as

```
micro config set --secret
```

Looks better than

```
micro config set --options.secret
```

(Config is a bad example here as it's a hardcoded one)

asim: When I say add to options I simply mean in the RPC requests. It's being added as an extra field but should mimic other protos that add an options field with Options message type that will contain the secret option. In regards to the CLI behaviour I agree --secret is better.

ben-toogood: I've amended the proto, waiting for builds to pass so we can merge.

ben-toogood (approved):

PR #1397, opened 2020-09-25
runtime: build services before running

ben-toogood:


profile/platform/platform.go
```diff
@@ -67,6 +70,20 @@ var Profile = &profile.Profile{
 			logger.Fatalf("Error configuring stream: %v", err)
 		}
 
+		// only configure the blob store for the store and runtime services
+		if ctx.Args().Get(1) == "runtime" || ctx.Args().Get(1) == "store" {
+			microStore.DefaultBlobStore, err = s3.NewBlobStore(
```
asim, line 75: Looking at this, I'm inclined to think if we were able to use something directly in k8s like etcd, nats, etc that it would be a better configuration. Because then we could offload the specifics to that install. Maybe we should just run the OSS minio and shift the config there?

proto/runtime/runtime.proto
```diff
@@ -151,3 +167,16 @@ message DeleteNamespaceRequest {
 }
 
 message DeleteNamespaceResponse {}
+
+
+message UploadRequest {
+	bytes data = 1;
```
asim, line 173: Should probably contain more identifying information about the source on the upload so that we're actually labelling it correctly on the server side. Right now its unclear to me from this how it storing the data on the server side but it should ideally be labelled by source name, version, etc.

proto/runtime/runtime.proto
```diff
@@ -151,3 +167,16 @@ message DeleteNamespaceRequest {
...
+message UploadRequest {
+	bytes data = 1;	
+}
+
+message UploadResponse {
+	string id = 1;
+}
+
+message BuildReadResponse {
+	bytes data = 1;
```
asim, line 181: What is the format of the response? if I'm being streamed a bunch of chunks of data, its unclear from this what it is. Either the request or the response needs to identify this clearly.

profile/platform/platform.go
ben-toogood: Yes, will action. I done this so we didn't need to pass the secrets into all the runtime deployments.

proto/runtime/runtime.proto
ben-toogood: It's streaming the binary back to the client. I'll add some comments.

proto/runtime/runtime.proto
```diff
@@ -168,9 +170,9 @@ message DeleteNamespaceRequest {
 
 message DeleteNamespaceResponse {}
 
-
 message UploadRequest {
 	bytes data = 1;	
+	Service service = 2;
```
asim, line 175: Best to rotate: Service = 1, data = 2

service/runtime/manager/manager.go
```diff
@@ -36,13 +37,34 @@ func (m *manager) Create(srv *runtime.Service, opts ...runtime.CreateOption) err
 		UpdatedAt: time.Now(),
 	}
 
-	// create the service in the underlying runtime
-	if err := m.createServiceInRuntime(service); err != nil && err != runtime.ErrAlreadyExists {
+	// if there is not a builder configured, start the service and then write it to the store
+	if builder.DefaultBuilder == nil {
```
crufter, line 41: Nice!

test/runtime_test.go
```diff
@@ -17,11 +16,6 @@ import (
 	"time"
 )
 
-const (
-	branch  = ""
-	version = "latest"
```
crufter, line 22: I would keep this, helps when wanting to run a non master version

crufter (approved):

test/runtime_test.go
ben-toogood: I don't think it's my PR which removed this, I think it's because i'm merging into a non-master branch which might be outdated.

cmd/platform/kubernetes/uninstall.sh
```diff
@@ -13,10 +13,6 @@ bash uninstall.sh;
 cd ../nats;
 bash uninstall.sh;
 
-# delete the PVs and PVCs
-kubectl delete pvc --all
-kubectl delete pv --all
```
domwong, line 18: Why don't we delete the PVs anymore?

ben-toogood: It took absolutely forever when testing locally. No real reason other than that.

PR #1413, opened 2020-10-01
Reintroducing config.Read for backwards compatibility reasons

crufter:


crufter:
cc @Dom so it looks like `config.Read` works. I haven't done `Delete` as deletes can be done from the CLI too. At that point users should really upgrade I think. Let me know however if it's needed.

Thanks

service/config/server/handler.go
```diff
@@ -99,6 +103,62 @@ func (c *Config) Get(ctx context.Context, req *pb.GetRequest, rsp *pb.GetRespons
...
+	buf := new(bytes.Buffer)
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	err = enc.Encode(dat)
+	if err != nil {
+		return merrors.BadRequest("config.Config.Read", "JSOn encode error: %v", err)
+	}
+	rsp.Change.ChangeSet.Data = strings.TrimSpace(buf.String())
+	rsp.Change.ChangeSet.Format = "json"
+	fmt.Println(string(rsp.Change.ChangeSet.Data))
```
domwong, line 158: Don't think we need this

domwong (approved):

PR #1417, opened 2020-10-01
Kubernetes Profile

ben-toogood:
Implement a working zero-dep kubernetes profile.


asim:
Can you add a doc in docs/ as part of this to explain profiles and how the kubernetes profile works

PR #1425, opened 2020-10-05
Remove source existence check from `micro update` 

crufter:
Fixes https://github.com/m3o/dev/issues/494

```sh
$ micro status
NAME	VERSION	SOURCE		STATUS	BUILD	UPDATED		METADATA
tags	latest	blog/tags	running	n/a	5h28m9s ago	owner=admin, group=micro
$ micro update tags
Service at github.com/micro/services/tags@master not found
```

We only have update tests for local code so this recent change was not picked up.
Adjusted a test to properly test for this.


ben-toogood (approved):

crufter:
The test fails which points out the fact that update does not work according to the reference https://m3o.org/reference#updating-a-service, seems like `micro update` parses `subfolder-branch` as a local folder.

client/cli/run/service.go
```diff
@@ -154,6 +154,22 @@ func appendSourceBase(ctx *cli.Context, workDir, source string) string {
 	domain := strings.Split(source, "/")[0]
 	_, err := publicsuffix.EffectiveTLDPlusOne(domain)
 	if !isLocal && err != nil {
+		// read the service. In case there is an existing service with the same name and version
```
crufter, line 157: Not pretty but it works

asim: easier to put it in a separate function so it just cleanly goes `if matchService && err == nil { return getSource(s) }`

crufter: Can do sure

PR #1438, opened 2020-10-07
auth/jwt: fix generation of keys (configure auth.DefaultAuth)

ben-toogood:
Previously the internal packages were auto-generating JWT private / public keys and then setting them using os.Setenv, now it directly configures the auth implementation. This solves a bug in the network where the auth package was used before the env vars are parsed in cmd/cmd.go.


go.sum
```diff
@@ -248,6 +248,7 @@ github.com/mattn/go-runewidth v0.0.7 h1:Ei8KR0497xHyKJPAv59M1dkC+rOZCMBJ+t3fZ+tw
 github.com/mattn/go-runewidth v0.0.7/go.mod h1:H031xJmbD/WCDINGzjvQ9THkh0rPKHF+m2gUSrubnMI=
 github.com/mattn/go-tty v0.0.0-20180219170247-931426f7535a/go.mod h1:XPvLUNfbS4fJH25nqRHfWLMa1ONC8Amw+mIA639KxkE=
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
+github.com/micro/go-micro v1.18.0 h1:gP70EZVHpJuUIT0YWth192JmlIci+qMOEByHm83XE9E=
```
crufter, line 251: go mod tidy should fix this

crufter (approved):

PR #1441, opened 2020-10-08
Api "service auth: route not found, unauthorized request" fix 

crufter:
In staging and prod, when calling the API from a non micro namespace, we have the following issue:

```sh
{"Id":"go.micro.client","Code":500,"Detail":"{\"Id\":\"go.micro.client\",\"Code\":500,\"Detail\":\"service auth: route not found\",\"Status\":\"Internal Server Error\"}","Status":"Internal Server Error"}
        unauthorized request
```

Please note that the above is a literal response body and it's not a copypaste error. It's not even a valid JSON.

- [x] Reproduce in tests
- [x] Fix


service/api/auth/wrapper.go
```diff
@@ -125,6 +125,7 @@ func (a authWrapper) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	} else if err != auth.ErrForbidden {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
```
crufter, line 128: This was missing, which caused the mangled output

service/auth/client/client.go
```diff
@@ -139,7 +139,7 @@ func (s *srv) Rules(opts ...auth.RulesOption) ([]*auth.Rule, error) {
 	}
 
 	callOpts := append(s.callOpts(), cache.CallExpiry(time.Second*30))
-	rsp, err := s.rules.List(options.Context, &pb.ListRequest{
+	rsp, err := s.rules.List(context.DefaultContext, &pb.ListRequest{
```
crufter, line 142: This fixes the `auth service not found` issue.

ben-toogood (approved):

PR #1445, opened 2020-10-09
Prevent generation of default accounts

crufter:
Notable changes
- Only generate default account in default namespace
- Added `DisableAdmin` flag to even disable this behaviour. This is useful for security hardening purposes, ie. imagine data loss occuring and then an attacker being able to log in with default credentials.
- Added `--namespace` option to `micro auth account create` as test which previously depended on account being autogenerated after namespace change now must create an admin account in that namespace first.

- [x] Fix
- [x] Test


cmd/cmd.go
```diff
@@ -126,6 +126,11 @@ var (
 			EnvVars: []string{"MICRO_AUTH_PRIVATE_KEY"},
 			Usage:   "Private key for JWT auth (base64 encoded PEM)",
 		},
+		&cli.BoolFlag{
+			Name:    "no_default_account",
```
asim, line 130: --disable_admin / MICRO_AUTH_DISABLE_ADMIN

cmd/cmd.go
```diff
@@ -127,6 +127,11 @@ var (
 			EnvVars: []string{"MICRO_AUTH_PRIVATE_KEY"},
 			Usage:   "Private key for JWT auth (base64 encoded PEM)",
 		},
+		&cli.BoolFlag{
```
ben-toogood, line 130: Please add this as a flag in service/auth

service/auth/server/auth/auth.go
```diff
@@ -59,6 +61,12 @@ func (a *Auth) Init(opts ...auth.Option) {
 }
 
 func (a *Auth) setupDefaultAccount(ns string) error {
+	if ns != namespace.DefaultNamespace {
```
ben-toogood, line 64: Why is this check required? Doesn't this circumvent the need for DisableAdmin flag?

ben-toogood (changes requested):

service/auth/server/auth/auth.go
crufter: No. Admins from now on only get generated for the default account.

cmd/cmd.go
crufter: Why? I don't see the `MICRO_AUTH_PRIVATE_KEY` flags there and it's a configuration option just like `MICRO_AUTH_PRIVATE_KEY`?

ben-toogood: It is specific to the auth server, MICRO_AUTH_PRIVATE_KEY and MICRO_AUTH_PRIVATE_KEY are used by all core components.

service/auth/server/auth/auth.go
ben-toogood: So if DisableAdmin is set, how do we ever create the first root account?

crufter: We don't. See ticket description

> Added DisableAdmin flag to even disable this behaviour. This is useful for security hardening purposes, ie. imagine data loss occuring and then an attacker being able to log in with default credentials.

cmd/cmd.go
crufter: Makes sense, thanks

ben-toogood (approved):

PR #1470, opened 2020-10-13
Remove dependancy on go-micro/v3/metadata package

ben-toogood:


asim:
One note on this PR. I've wondered whether the Metadata type should actually be scoped under the service/context package. So next step might be to move it there.

ben-toogood:
> One note on this PR. I've wondered whether the Metadata type should actually be scoped under the service/context package. So next step might be to move it there.

I'll action this in this PR.

asim:
I mostly meant context.Metadata rather than context/metadata.Metadata but we can come back to it.

PR #1480, opened 2020-10-14
Remove dependancy on go-micro/v3/selector package

ben-toogood:


asim:
Vasiliy made a comment that this should be part of the router and I sort of now agree but the refactor will take work. Let's look at it after.

ben-toogood:
Yeah, I completely agree, they're too linked to be standalone packages.

PR #1499, opened 2020-10-16
Move all service cli commands to service/cli package

asim:
The service package is a library now, so it requires cleanup. All cli related stuff should be in a single package.


asim:
Closing in favour of moving to client/cli because these appear to be cli commands not service startup commands

PR #1505, opened 2020-10-16
auth: fix error token expired

ben-toogood:
Improve the fallback logic when a services refresh token expires (e.g. in cases where the auth service is temp unavailable). When this happens, the service will now generate a new refresh/access token using the credentials injected by the runtime manager.


crufter (approved):

cmd/util.go
```diff
@@ -148,9 +148,19 @@ func refreshAuthToken() {
 			// generate the first token
 			tok, err := auth.Token(
 				auth.WithToken(tok.RefreshToken),
-				auth.WithExpiry(time.Minute*10),
+				auth.WithExpiry(time.Second*10),
```
domwong: Why switch to 10s?

ben-toogood: Mistake, fixed

PR #1516, opened 2020-10-19
runtime: vendor dependencies 

ben-toogood:
Closes https://github.com/m3o/dev/issues/469


asim (approved):

client/cli/run/service.go
```diff
@@ -213,6 +214,20 @@ func runService(ctx *cli.Context) error {
...
+		// after we finish the upload
+		vendorDir := filepath.Join(source.LocalRepoRoot, "vendor")
+		if _, err := os.Stat(vendorDir); os.IsNotExist(err) {
+			defer os.RemoveAll(vendorDir)
+		} else if err != nil {
+			return err
+		}
+
+		// vendor the dependencies
+		if err := run.VendorDependencies(source.LocalRepoRoot); err != nil {
```
domwong, line 227: I wonder if we should move this so it only attempts to vendor if the vendor dir is not there i.e. if user has a vendor dir we respect and use that instead. https://golang.org/ref/mod#go-mod-vendor 
> Note that go mod vendor removes the vendor directory if it exists before re-constructing it. Local changes should not be made to vendored packages. The go command does not check that packages in the vendor directory have not been modified, but one can verify the integrity of the vendor directory by running go mod vendor and checking that no changes were made.

ben-toogood: Right now if a user already has a vendor directory it'll just update that, which I think is okay?

service/runtime/local/local.go
```diff
@@ -128,7 +128,13 @@ func (r *localRuntime) Create(resource runtime.Resource, opts ...runtime.CreateO
 		}
 		if len(options.Command) == 0 {
 			options.Command = []string{"go"}
-			options.Args = []string{"run", "."}
+
+			// not all source will have a vendor directory (e.g. source pulled from a git remote)
```
domwong, line 132: Don't think this comment is accurate (not that it makes a difference to the code) - people do commit the vendor dir to git

ben-toogood: What I mean by this is if the source was pulled from a git remote, the CLI hasn't had the chance to vendor the deps so whilst the vendor dir may have been committed, most of the time it won't have been.

client/cli/run/service.go
domwong: From the link I pasted it doesn't seem to update, just blow away the vendor dir and recreate. One edge case i can think of off the top of my head is some people, perhaps incorrectly, might make changes inside their vendor directory to tweak the behaviour. Running `go mod vendor` will blat any changes they've made.

@asim any strong feelings on how the vendoring should behave?  Should we respect the user's existing vendor dir? Or re-vendor it ourselves always?

PR #1523, opened 2020-10-21
Allows the usage of gRPC+Web whenever gRPC is being used

arbarlow:
In this particular case, it's mostly to allow usage in the proxy
but since it's implemented here, it allows for usage I think elsewhere
too!

Part of https://github.com/m3o/dev/issues/517


asim:
LGTM

asim (approved):

PR #1545, opened 2020-10-28
Upgrade Guide

ben-toogood:


asim:
Comments to action

```
Cool. Perhaps then just a note to head off the inevitable question of what happens if I still want to use dotted names. Will save us answering the same question over and again in Slack.
```
https://github.com/m3o/dev/pull/543/files#r512803015

```
I wonder if we should call out the move from go-micro to micro. Something like - "As part of v3 we've simplified the development model and consolidated the micro and go-micro libraries. This means that you, as the developer, only need to worry about leveraging the libraries offered in micro"
```
https://github.com/m3o/dev/pull/543/files#r512527087

Once these are done, we can merge and then continue to add as needed.

PR #1552, opened 2020-10-30
Allow Server options to specific gRPCWeb options and boot if added

arbarlow:
Allow Server options to specific gRPCWeb options and boot if added, I then set these in the proxy and use the next port number.

I wasn't sure on the direction here as it includes gRPCWeb specific options in the server options interface, but happy to take direction if there's a better way!


asim:
I think what it leads us to is the rebirth of micro web, starting solely with grpc-web support and eventually moving into some sort of rendering component. 

Let's move this into micro/service/web as its own thing, then web.Address and web.Options are a thing and micro/service/cli can be augmented to include a `micro service web` command to spin up this thing on port :8082. We can reuse as much of the code as possible, etc.

asim:
As part of this or perhaps a separate PR (probably separate), can you generate the typescript client for all our top level protos in micro/client/web using some script we can then automate with github actions (much like we do with client/sdk). Then we can publish the typescript client.

asim:
Considering our 3.0 release is happening on Thursday, I think we have to hold off on merging anything grpc-web related that isn't the final thing that's going to be used.

My initial assumption was, seamless support into the proxy would be a quick win. Running on a separate port now augments micro in a way I think we have to sort of revisit the overall architecture and experience. Everything "web" related used to run through `micro web` including http proxying to web apps. If we're leaning in the direction of any sort of grpc-web support that then starts to factor in a web experience once more. It is not out of the question to resurrect micro web. Many people were annoyed it was removed and there's something to be said for built in UI based querying. Considering most open source provides some out of the box functionality we may need to do it.

I think we should open a new doc to address to address this. I will cc @domwong to make sure we get a doc done first and hold off on merging anything before the release. At the very least if we're going to have grpc-web support that runs on a separate port from the proxy its going to be a totally separate service since there's logical separation of concerns here and that's what micro provides codified in service form.

PR #1554, opened 2020-10-30
Deploy nginx

ben-toogood:
• Use nginx ingress controller in the platform deployment. 
• TLS is terminated at nginx. 
• nginx uses CF DNS resolver to generate certs


asim:
LGTM. Next step, remove CF/certmagic (but leave tls and autocert support).

PR #1557, opened 2020-11-02
Fix : cmd options flags

crazybber:
allow flags and action in cmd


ben-toogood:
Can you please add a description to the PR outlining why it is required

crazybber:
developer need pre-parsed flags to init service parameters ,so we can do it in a temp cmd.New() with inside service.New().

PR #1558, opened 2020-11-02
Fix windows lock sharing violation

crazybber:
mtx.Lock() conflict with lock.Lock() on windows.

this is just ` Get(path string) `  :a reading operation , no need using double lock for reading race in single process.

Or we can remove mtx.Lock() to allow muti-process to read ,but still use file lock to handle race by syscal.


crazybber:
@asim @ben-toogood

PR #1562, opened 2020-11-04
[WIP] Windows CI

crufter:
Ignore this for now pls.

### DIFF
--- a/.github/workflows/integration-win.yml
+++ b/.github/workflows/integration-win.yml
@@ -1,0 +1,95 @@
+name: Windows CI Profile
+on: [push]
+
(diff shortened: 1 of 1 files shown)


client/cli/init/init.go
```diff
@@ -50,7 +50,7 @@ func Run(ctx *cli.Context) error {
 
 	// write the profiles
 	for _, profile := range profiles {
-		path := filepath.Join(Import, profile, Version)
```
crufter, line 53: This was causing `import github.com\micro\micro` to be generated in profile.go and make compilation fail.

PR #1569, opened 2020-11-06
profile: use ~/.micro/server for local filestore

ben-toogood:
Fixes an issue where the local accounts (persisted in the store) would disappear upon restart.

### DIFF
--- a/profile/profile.go
+++ b/profile/profile.go
@@ -92,7 +92,7 @@ var Local = &Profile{
 	Name: "local",
 	Setup: func(ctx *cli.Context) error {
 		microAuth.DefaultAuth = jwt.NewAuth()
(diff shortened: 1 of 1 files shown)


profile/profile.go
```diff
@@ -91,7 +92,7 @@ var Local = &Profile{
 	Name: "local",
 	Setup: func(ctx *cli.Context) error {
 		microAuth.DefaultAuth = jwt.NewAuth()
-		microStore.DefaultStore = file.NewStore()
+		microStore.DefaultStore = file.NewStore(file.WithDir(filepath.Join(user.Dir, "server")))
```
crufter, line 95: Perhaps `server/store`? I think all folders should move under `server` ie logs uploads etc but I appreciate that's a much harder fix. Maybe to keep the same spirit `server/store` should  do for now.

crufter (approved):

PR #1573, opened 2020-11-09
Add the ability to override the micro dir (ie. what is now `~/.micro`)

crufter:


profile/profile.go
```diff
@@ -213,11 +213,14 @@ func SetupJWT(ctx *cli.Context) {
 }
 
 func SetupConfigSecretKey(ctx *cli.Context) {
+	// set the micro dir location
+	user.SetDir(ctx.String("dir_path"))
```
crufter, line 217: Maybe rename SetDir to Init?

crufter:
Parked for now as I don't need it anymore but will revisit it later as it's a useful feature in itself.

PR #1575, opened 2020-11-10
fix the alignment @ https://micro.mu/reference\#rules

lambdaR:
Replaced hard tag(s) with space character(s) instead.
Markdown rule [MD010](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md010)


asim:
Thanks for the contribution

PR #1579, opened 2020-11-12
some changes to Dockerfile to make micro docker image more convenient for use

lambdaR:
some changes made to the Dockerfile:

1. install musl-libc variant of protoc `RUN apk add protobuf` [package](https://pkgs.alpinelinux.org/package/edge/main/x86/protobuf) to avoid **no such file or directory** error after you install the protoc [glib](https://github.com/protocolbuffers/protobuf/releases) variant.
2. include /root/bin to PATH `ENV PATH="/root/go/bin:${PATH}"`  to avoid [Errors](https://github.com/micro/micro/tree/master/cmd/protoc-gen-micro#errors)
3. move micro to /usr/bin `COPY --from=builder /micro /usr/bin` so you can run micro globally.
4. add server parameter `ENTRYPOINT ["micro", "server"]` so you can create a container via `docker run -d --name micro -p 8080:8080 micro/micro` instead of `docker run -d --name micro -p 8080:8080 micro/micro server`


crufter (approved):

asim:
I merged but reverted. I think this is going to cause some problems with our usage because we have `micro service [x]` commands and `micro cli` and other things we want to preserve.

PR #1593, opened 2020-11-19
Fixes to resource quota and network policy

domwong:
Update for these resources was broken. Have fixed and added tests to cover our internal k8s client for create/update/delete actions.


ben-toogood (approved):
Nice improvement. It would be good to move away from our own k8s client at some point soon, I don't see any need for us to have our own when there are packages such as: https://github.com/kubernetes/client-go.

domwong:
> Nice improvement. It would be good to move away from our own k8s client at some point soon, I don't see any need for us to have our own when there are packages such as: https://github.com/kubernetes/client-go.

1000% agree

PR #1607, opened 2020-11-24
OpenAPI generator for Micro service protos

chrusty:
This is a `protoc` plugin to generate OpenAPI spec (as JSON).

- Creates schema components for each proto message it finds
- Creates path components for each RPC service method it finds, along with the associated request & response payloads
- Includes server objects for the dev and prod Micro platforms (though I guess these will be consolidated soon)
- Supports Micro's authentication mechanism (API token)
- Supports the namespace header
- Some tests (though these require the `protoc` binary)


cmd/protoc-gen-openapi/main.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+const (
+	microLogLevelEnvVar = "MICRO_LOG_LEVEL"
+)
+
+func main() {
+
+	// Make a Logrus logger:
+	// The Micro logger is unable / unwilling to output to os.Stderr (code generators break if logs go to os.Stdout)
```
domwong, line 20: will look at fixing this

domwong (approved):

PR #1612, opened 2020-11-26
Pre-generated OpenAPI specs for 4 Micro services

chrusty:
Here are OpenAPI specs & docs for some of the core Micro services:
- [x] Auth
- [x] Broker
- [x] Config
- [x] Events
- [x] Runtime
- [x] Signup
- [x] Store


asim:
Cool, can you scope to client/api dir

PR #1630, opened 2020-12-08
simplify log reading and add sync.WaitGroup for streaming

asim:


service/runtime/kubernetes/kubernetes.go
```diff
@@ -80,28 +78,38 @@ func (k *kubernetes) Logs(resource runtime.Resource, options ...runtime.LogsOpti
...
 			if err != nil {
 				logger.Errorf("Failed to get logs for service '%v' from k8s: %v", s.Name, err)
 				return nil, err
 			}
+
+			// create a stream even though we're not streaming
 			kstream := &kubeStream{
-				stream: make(chan runtime.Log),
+				// make a stream buffer of size records
+				stream: make(chan runtime.Log, len(records)),
```
crufter, line 92: So you nuked the goroutine because the channel was made async right?

crufter (approved):

service/runtime/kubernetes/kubernetes.go
asim: yes

PR #1638, opened 2020-12-13
OpenTracing

chrusty:
Since we're looking at tracing it makes sense to do this the proper way: OpenTracing (OpenTelemetry)

- [x] Added an OpenTelemetry package, which exposes a DefaultTracer
- [x] Added a Jaeger implementation which can be used to set the DefaultTracer
- [x] By default the Jaeger implementation will only trace if that decision has already been made, and propagated through contex
- [x] Optional sampling rate
- [x] Added a new Wrapper for OpenTracing which uses the DefaultTracer to instrument RPCs
- [x] Split up the rest of the wrappers into individual files whlie I was in there
- [x] Use profile to set the default OpenTracer where appropriate
- [x] Figure out how to set the initial trace decision in a client (```ext.SamplingPriority.Set(span, 1)```)
- [x] Modify the default server to wrap RPCs with the OpenTracing wrapper
- [x] Hook a parent span into the API handler so any micro API call gets traced (later we can make this a percentage)
- [ ] Include Jaeger in the stack (standalone pod for now is OK)
- [ ] Add configuration for the jaeger address in each env
- [ ] Figure out why context doesn't seem to get passed in early-stage API handling


internal/opentelemetry/http_wrapper.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// Concatenate the operation name:
+	operationName := fmt.Sprintf("API.%s", r.URL.RawPath)
+
+	// Start a span:
+	span, newCtx := opentracing.StartSpanFromContextWithTracer(r.Context(), DefaultOpenTracer, operationName)
+	ext.SamplingPriority.Set(span, 1)
+	defer span.Finish()
+
+	logger.Warnf("OpenTelemetry Wrapping (%s)", reflect.TypeOf(DefaultOpenTracer))
```
chrusty, line 27: @asim Running this wrapper in the micro API, always coming up with the default (NoopTracer), like the profile hasn't set the correct tracer. I'm guessing there is something different with the profile interaction for API.

asim: What is DefaultOpenTracer here? If you're looking at service/debug then it gets initialised at the time of service start.

asim: https://github.com/micro/micro/blob/master/service/debug/debug.go#L15

chrusty: OK I've just taken control of this in the API service now, seems to work.

kunaldawn:
any plans to merge this?

chrusty:
> any plans to merge this?

@kunaldawn yes, hopefully soon!

chrusty:
@kunaldawn please see the new PR for this: https://github.com/micro/micro/pull/1805

PR #1663, opened 2021-01-12
Add @crufter's model

asim:
Merging @crufter's work on the model. Includes changes from this PR to the interface https://github.com/micro/dev/pull/11. Will continue to evolve from here and potentially include further future models like sql, graph, cache, etc.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/gobwas/httphead v0.0.0-20180130184737-2c6c146eadee
 	github.com/gobwas/pool v0.2.1 // indirect
 	github.com/gobwas/ws v1.0.3
+	github.com/gofrs/uuid v3.2.0+incompatible
 	github.com/golang/protobuf v1.4.3
 	github.com/google/uuid v1.1.2
 	github.com/gorilla/handlers v1.4.2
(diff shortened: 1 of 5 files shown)


crazybber:
this is quite cool

crazybber:
i think if we can add grpc reflection ablity into server end to  support some case ,such as grpcui(like postman  but for grpc.)

asim:
> this is quite cool
> 
> i think if we can add grpc reflection ablity into server end to support some case ,such as grpcui(like postman but for grpc.) with enviroment flag support ,such as a debug mode.

If we change the way in which we register handlers then thats possible. Right now its difficult because we have our own custom grpc server implementation.

crazybber:
> > this is quite cool
> > i think if we can add grpc reflection ablity into server end to support some case ,such as grpcui(like postman but for grpc.) with enviroment flag support ,such as a debug mode.
> 
> If we change the way in which we register handlers then thats possible. Right now its difficult because we have our own custom grpc server implementation.

use a static router to by pass 3rd register ability ,just provide direct gprc service call by name plus with grpcui/postman ,which is quite useful and efficiency in development enviroment  for developing.

PR #1664, opened 2021-01-13
clean code : remove duplicate imports

crazybber:
clean code : remove following duplicate import:
 "github.com/micro/micro/v3/service/registry" 
 "github.com/micro/micro/v3/service/store" 
 "github.com/micro/micro/v3/service/router"


crazybber:
test failed ，same  with your merge https://github.com/micro/micro/runs/1695434875 ,which look we need to fix some test case

PR #1666, opened 2021-01-15
Model: check if store is nil

crufter:


service/model/model.go
```diff
@@ -334,6 +344,9 @@ func (d *model) List(query Query, resultSlicePointer interface{}) error {
 			if d.options.Debug {
 				fmt.Printf("Listing key '%v'\n", k)
 			}
+			if d.store == nil {
+				return ErrorStoreIsNil
+			}
```
keonjeo, line 349:
```golang
if d.store == nil {
    return ErrorStoreIsNil
}
```

Why we could NOT move this block code into store.Read, store.Write, store.Delete method？

service/model/model.go
```diff
@@ -586,6 +599,9 @@ func (d *model) Delete(query Query) error {
 		if d.options.Debug {
 			fmt.Printf("Deleting key '%v'\n", key)
 		}
+		if d.store == nil {
+			return ErrorStoreIsNil
+		}
```
keonjeo, line 604: same above.

service/model/model.go
```diff
@@ -308,6 +315,9 @@ func (d *model) Read(query Query, resultPointer interface{}) error {
 			if d.options.Debug {
 				fmt.Printf("Listing key '%v'\n", k)
 			}
+			if d.store == nil {
+				return ErrorStoreIsNil
+			}
```
keonjeo, line 320: Why we could NOT move this block code into store.Read, store.Write, store.Delete method？

service/model/model.go
```diff
@@ -252,6 +256,9 @@ func (d *model) Create(instance interface{}) error {
 		if d.options.Debug {
 			fmt.Printf("Saving key '%v', value: '%v'\n", k, string(js))
 		}
+		if d.store == nil {
+			return ErrorStoreIsNil
+		}
```
keonjeo, line 261: Why we could NOT move this block code into store.Read, store.Write, store.Delete method？

service/model/model.go
```diff
@@ -243,6 +244,9 @@ func (d *model) Create(instance interface{}) error {
 			oldEntry != nil &&
 			!reflect.DeepEqual(getFieldValue(oldEntry, index.FieldName), getFieldValue(instance, index.FieldName)) {
 			k := d.indexToKey(index, id, oldEntry, true)
+			if d.store == nil {
+				return ErrorStoreIsNil
+			}
```
keonjeo, line 249: Why we could NOT move this block code into store.Read, store.Write, store.Delete method？

PR #1677, opened 2021-01-21
Add context to the model interface

asim:
Add context to the model interface so we can later use accounts from the context to set which database/table to query.

### DIFF
--- a/service/model/model.go
+++ b/service/model/model.go
@@ -6,6 +6,7 @@
 package model
 
 import (
+	"context"
 	"encoding/base32"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


crufter:
Could we use the following pattern:

Let's have a Context method which actually returns an entirely new instance of Model (method chaining) ie
```
Context(ctx context.Context) Model
```

So this way we dont pollute the API for users who don't care about contexts. Not perfect as method chaining is not part of the library yet but perfect time to introduce it.

asim:
Changed the PR to match what we discussed. The better pattern is to specify Context to the interface which then provides an instance with that initialised. Hopefully we can carry this pattern over to other interfaces.

crufter (approved):

PR #1704, opened 2021-02-04
Cross namespace auth rules

domwong:
This PR adds auth rules that are applied to accounts across namespaces. 

The current auth rules apply in the following situations
- public rules - these rules are applied to all callers, regardless of whether they pass an auth token or not
- account rules - these rules are applied to callers that pass a valid auth token for the namespace. If you pass an auth token issued by a different namespace it is ignored, and the call is treated as an unauthenticated one

This PR adds a further type of rule which is applied when a valid auth token is passed, regardless of which namespace it was issued by.  

Note: This is the danger zone™ because anyone who has a namespace can create an account issued by their namespace with whatever scopes they wish so it doesn't really give you any guarantee about the caller apart from that they have been provisioned by a namespace in micro. So these auth rules should only be used to allow access to endpoints that either require no protection or have been adequately protected which means checking the issuer and the scopes of the caller and correctly isolating callers from different namespaces. 

To create this new type of rule you use the special scope of `+`

```
micro auth create rule  --resource="service:helloworld:*" --priority 1 --scope "+" helloworld-public
```


asim:
Do you think there is another term for "cross issuer" or whether it's a rule "type". Or even if it's a type of scope.

domwong:
> Do you think there is another term for "cross issuer" or whether it's a rule "type". Or even if it's a type of scope.

Yeah I wasn't massively happy with it when I made it `crossissuer`. Thinking about it I think since we're already doing this overloading of the scope field it makes sense to do it there rather than having to add a `type` of rule. So perhaps the answer is 
- `<blank>` means public
- `*` means any scope issued by this namespace
- `+` means any scope from any namespace

asim:
Works for me and then I guess in future potential doing things like namespace:level e.g micro:admin

domwong:
Code and PR description updated to use `+`

asim (approved):

PR #1710, opened 2021-02-14
Replace boltdb/bolt with etcd/bbolt

Clivern:
This PR replaces the archived https://github.com/boltdb/bolt with https://github.com/etcd-io/bbolt. I see it is already replaced somewhere else inside micro `service/store/file`


asim:
There was a merge conflict, so closing in favour of https://github.com/micro/micro/pull/1759/files

PR #1722, opened 2021-03-03
Enable change secret without old password for micro admins, services

crufter:
Enables, without the knowledge of a user's password:
- micro admins changing passwords of users in other namespaces
- services changing passwords for users in their own namespace


asim (approved):

lambdaR:
@crufter @asim maybe we need to set a session timeout for logged in **admins**, people forget to log out manually and if a **gentleman** hacker managed to remotely ssh the server he/she will be able to change the password easily and takes control of micro.

PR #1743, opened 2021-03-25
remove single quote from call's and stream's usage phrases.

lambdaR:
currently, the `micro call` usage statement shows an example of how to use it:

`micro call greeter Say.Hello '{"name": "John"}' 
`

which leads to the following err:

**_Error creating request invalid character '\\'' looking for beginning of value_**


asim:
I'm making this call in production with quotes and it works and without quotes it returns

```
$ micro call helloworld Helloworld.Call {"name": "john"}
Error creating request invalid character 'n' looking for beginning of object key string
```

So this may be an environmental issue or shell issue

lambdaR:
interesting ...
123temp

lambdaR:
am using **zsh** Ubuntu terminal

asim:
Ah, you're using the interactive CLI. This is the issue. The example is for bash shell, not the micro cli.

lambdaR:
well, that is confusing .... one should stick to one tool (shell | cli). 
examples in usage should reflect the the tool you are using. 

for shell ==> `micro call helloworld Helloworld.Call '{"name":"foo"}'`
and for cli ==> `call helloworld Helloworld.Call {"name":"foo"}`

asim:
Nowhere do we promote the use of the interpreted cli in the readme

lambdaR:
ok ... closing

PR #1776, opened 2021-05-10
rename cockroach plugin to postgres

asim:


asim:
> @asim you probably already know that but Postgres does not support queries of the of type "CREATE DATABASE IF NOT EXIST".
> 
> We ended up forking the plug-in to make this work with Postgres 13 but all it take is amend some of the statements to handle database creation in a slightly different way.

Well that sucks... thanks for letting me know. I won't merge this until we have a better solution then. We're personally potentially moving off cockroach to a managed sql so I think having a generic postgres plugin makes sense. We also may expose that raw connection via  `DB(ctx context.Context) (*sql.DB, error)` method.

crashdump:
Other than the issue of the "NOT" statement being problematic, I can vouch for the rest of the code, it's been working great for us on PG13. 

We did not want to manage our database ourselves either.

crashdump:
Unfortunately nothing that will help you; I would have loved to contribute something useful to the project but we've simply removed the database management from the plug-in and passed the responsibility to a our infrastructure as code for now.

asim:
Well in light of that, I'm going to merge this and we'll see whether those things get removed. I think because the store requires a specific schema it makes sense for us to create it but equally as you mention from a security perspective its better to do this as a separate piece of work.

crashdump:
Even better, this should work nicely.

asim:
Cheers, all changes merged so that should be good to go.

crashdump:
Something else I had forgotten about... the table name format `micro.thing` is interpreted by Postgres as reference to a schema (see [https://www.postgresql.org/docs/9.1/ddl-schemas.html](https://www.postgresql.org/docs/9.1/ddl-schemas.html).

Once I renamed the table+database concatenation from `micro.thing` to `micro_thing`, everything went well. See [https://github.com/crashdump/micro/commit/ce748cfa03203272b62e7927a00dabd57df7aac6](https://github.com/crashdump/micro/commit/ce748cfa03203272b62e7927a00dabd57df7aac6)

I've not raised a PR with my changes as it would be a breaking change, so I'll leave that with you.

asim:
I think the thing to note here though. Isn't that just going to create tables of name database_table. Where as what we actually want is tables under a specific database.

crashdump:
What was I smoking – you are totally right. I did solve that correctly on my fork by removing the reference to the database which is unnecessary with Postgres. In fact `%s.%s -> database, table` can be replaced by `%s -> table`. 

I don't know if this works with Cockroach though.

asim:
I think the one issue here is, our store can connect to multiple databases using a single sql connection and by stripping this reference we lose that functionality entirely. So the straight swap becomes pretty complicated. I think I have to admit defeat and back this out until I can spend the proper time on it.

crashdump:
Not all is lost! I've done further testing using schemas with Postgres, while keeping the existing behaviour with Cockroach (multiple databases).

`var version string
if err = db.QueryRow("select version()").Scan(&version); err != nil {
	if strings.Contains(version, "PostgreSQL") {
		_, err = db.Exec(fmt.Sprintf("CREATE SCHEMA IF NOT EXISTS %s;", database))
		if err != nil {
			return err
		}
	}
}`

It seems to do the trick, see https://github.com/crashdump/micro/commit/1f1cfd13e702d435e2a930b2f6a5115fc0d0c540

I've tested with Cockroach and Postgres 13. Tests could do with a little bit more depth (such as testing the keys and databases are in the right place) but manual testing looks good.

asim:
That looks like a nice clean version we could potentially use in a forward rolling fashion. I'll need to do some tests as well just look and see how it all is.

crashdump:
Happy to report the postgres schema solution works fine.

asim:
Great to hear. I think we'll be migrating to managed postgres soon so likely that's the path.

PR #1799, opened 2021-05-20
Add v3 cell variant that has headless chrome installed

crufter:


asim:
Colocated with service

PR #1805, opened 2021-05-25
OpenTelemetry

chrusty:
Replacing a previous PR (the repo had changed too much in the interim).

- [x] Added an OpenTelemetry package, which exposes a DefaultTracer
- [x] Added a Jaeger implementation which can be used to set the DefaultTracer
- [x] By default the Jaeger implementation will only trace if that decision has already been made, and propagated through contex
- [x] Optional sampling rate
- [x] Added a new Wrapper for OpenTracing which uses the DefaultTracer to instrument RPCs
- [x] Split up the rest of the wrappers into individual files whlie I was in there
- [x] Use profile to set the default OpenTracer where appropriate
- [x] Figure out how to set the initial trace decision in a client (```ext.SamplingPriority.Set(span, 1)```)
- [x] Modify the default server to wrap RPCs with the OpenTracing wrapper
- [x] Hook a parent span into the API handler so any micro API call gets traced (later we can make this a percentage)

Todo:
- [ ] Include Jaeger in the stack (standalone pod for now is OK)
- [ ] Add configuration for the jaeger address in each env


domwong:
Looks good. 

@asim how does this fit with the existing debug stuff? I think there's tracing in there or something? Should this replace it? Do we need to rip something out?

asim:
I think we just ignore what exists for now. I'm treating Debug.{Health, Stats, Trace, Log, ...} as some built in stuff for internal usage but with something like this its going to give us out of the box tracing using Jaeger or whatever else which will be super useful. Jaeger's UI can actually be embedded so we can drop it into the m3o/cloud dashboard if we want to.

profile/profile.go
```diff
@@ -175,6 +191,21 @@ var Kubernetes = &Profile{
...
 		client.DefaultClient.Init(client.Router(router.DefaultRouter))
+
+		// Configure tracing with Jaeger:
+		tracingServiceName := ctx.Args().Get(1)
+		if len(tracingServiceName) == 0 {
+			tracingServiceName = "Micro"
+		}
+		openTracer, _, err := jaeger.New(
+			opentelemetry.WithServiceName(tracingServiceName),
+			opentelemetry.WithTraceReporterAddress("jaeger:6831"),
```
asim, line 202: What happens if this is not running? The helm chart does not start jaegar

chrusty: Yes, we need Jaeger.

Jaeger itself operates with UDP, so it doesn't matter if Jaeger is just not _accessible_. The only problem is the hostname - if it fails to resolve the configured hostname then this part will fail.

So one approach is to just default to `localhost` (because we can always resolve that), then override that config if we have Jaeger running.

asim: That works. I think this k8s profile just doesn't have any deps because there's nothing running on k8s basically. It's the m3o/platform profile where we'll probably inevitability end up configuring whatever we run and wherever it lives.

PR #1827, opened 2021-08-04
Add content type to blob write options

asim:


plugin/s3/s3.go
```diff
@@ -151,10 +152,11 @@ func (s *s3) Write(key string, blob io.Reader, opts ...store.BlobOption) error {
...
 	object := sthree.PutObjectInput{
-		Bucket: &s.options.Bucket,
-		Key:    &k,
-		Body:   strings.NewReader(buf.String()),
-		ACL:    aws.String(acl),
+		Bucket:      &s.options.Bucket,
+		Key:         &k,
+		Body:        strings.NewReader(buf.String()),
+		ACL:         aws.String(acl),
+		ContentType: &s.options.ContentType,
```
asim, line 159: Should be options.ContentType

PR #1830, opened 2021-08-09
Turn off html escaping when json encoding in the api

crufter:
See POC https://play.golang.org/p/oBNxUjVTzus


domwong (approved):
:shipit: 

Fix the tests (obvs) but also need some testing to make sure it doesn't break the many layers that we have API -> network -> anything else

PR #1860, opened 2021-11-05
multipart form data support

domwong:
Adding support for `multipart/form-data` calls to the API. 

Allows us to then do something like this

`curl api.m3o.com/foo/bar -F hello=world -F file=@/foo.jpg` 

Gets translated and sent to a service with a request that looks like
```
message BarRequest {
  string hello = 1;
  bytes file = 2;
}
```

This works in two steps. 

First we now support content type of `mutipart/form-data` and take values from that and map to a json. Form fields are strings (regardless of whether they should be a bool, int, etc). This gets sent over the wire. 

When it hits the target service we check the content type of the incoming message, if it's a json codec we will try to parse the message in the normal way and if we fail we will try to parse the json again in a more permissive style - converting strings in to the correct target type (bool, int, etc).

### DIFF
--- a/service/api/api.go
+++ b/service/api/api.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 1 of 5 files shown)


domwong:
Tested in dev env as part of this https://github.com/micro/services/pull/264

PR #1872, opened 2021-12-29
[FEATURE] Micro run with Live-Reloading

h1z3y3:
We had a discussion before here: https://github.com/micro/services/discussions/345
It's an useful feature for micro service developer.

Add two args for `micro run`:
1. `--watching`, integer, specific the detect delay for file changes
2. `--force`, boolean, if it need rebuild and restart the running service


client/cli/run/cli.go
```diff
@@ -46,6 +46,15 @@ var flags = []cli.Flag{
 		Name:  "metadata",
 		Usage: "Set any metadata on the service e.g. foo=bar",
 	},
+	&cli.IntFlag{
+		Name:  "watching",
```
asim, line 50: Change to "watch"

client/cli/run/cli.go
```diff
@@ -46,6 +46,15 @@ var flags = []cli.Flag{
 		Name:  "metadata",
 		Usage: "Set any metadata on the service e.g. foo=bar",
 	},
+	&cli.IntFlag{
```
asim, line 49: Use Bool so it's --watch=true

client/cli/run/cli.go
```diff
@@ -46,6 +46,15 @@ var flags = []cli.Flag{
...
+		Name:  "watching",
+		Usage: `Watch the file changes of source directories, then rebuild and restart the service. 
+		e.g. watching=500 means watching delay time is 500ms, if watching=0 feature will be disabled.`,
```
asim, line 52: Maybe call the flag --watch_delay to separately define the delay

h1z3y3: Okay, use `--watch_delay` already

h1z3y3: Now we only watch the `*.go` and `*.proto` file and only support `delay` feature

We can add some other configurations for live-reloading in the future,
like `--watch_include` or `--watch_exclude`, etc.

But for now, I think  it's enough already.

client/cli/run/service.go
```diff
@@ -376,48 +423,9 @@ func runService(ctx *cli.Context) error {
 	// run the service
 	err = runtime.Create(srv, opts...)
 
-	watchDelay := ctx.Int("watching")
+	watchDelay := ctx.Int("watch_delay")
 	if source.Local && watchDelay > 0 {
```
asim, line 427: I would imagine the watcher runs if I specify --watch=true. Then if I need a watch delay I add the flag --watch_delay afterwards. So you're missing the watch flag.

h1z3y3:
@asim  `io/fs` was introduced since Go1.16, so the checks failed.

I replace it with package `os`, and using `filepath.Walk()` to replace `filepath.WalkDir()`, but it may by little ineffcient.

https://github.com/golang/go/blob/91e782106ea465acc6a4c719081cefb690b28533/src/path/filepath/path.go#L498-L511

image

asim:
Bump the go versions in pr.yaml and tests.yaml 

https://github.com/micro/micro/tree/master/.github/workflows

asim:
Thanks for the contribution. Looks like a great feature.

PR #1907, opened 2022-02-15
change error response to using lowercase for consistency response with success response based on standard JSON REST API

agus7fauzi:
In my company is currently using micro, we found irregularities when making a successful response and a failed response especially during input validation. For the response we use the json standard, so for the first letter is lowercase, we encounter problems when giving a failed response because the first letter of the failed response is an uppercase letter. we hope micro fixes this to be able to use first lower case (camelCase) for failed responses, thanks.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/micro/micro/v3
+module github.com/agus7fauzi/micro/v3
 
 go 1.16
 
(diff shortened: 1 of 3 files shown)


asim:
This is not the correct solution. The appropriate solution is to add json tags to the error struct.

asim:
type Errors struct {
  Id string `json:"id"`
  ....
}

You can also define your own errors.

agus7fauzi:
Hi, I have resubmitted the pull request according to the advice you gave, thank you, https://github.com/micro/micro/pull/1908

PR #1953, opened 2022-07-01
Easy install micro on windows using Scoop

a0v0:


asim:
Any reason why its not just called micro? If it can't be that then micro-cli rather than microcli.

a0v0:
ok let me change that

a0v0:
done please check now

PR #1960, opened 2022-08-17
add pgx plugin

Alexfilus:
Added plugin for micro KV store based on PostgreSQL and  popular pgx driver.


asim:
If we're largely replicating the postgres plugin for pgx is there any reason not to just switch the driver there? What's advantages of of pgx over the existing driver?

Alexfilus:
Maintainer of https://github.com/lib/pq#status had recommended to use https://github.com/jackc/pgx
It is faster https://devandchill.com/posts/2020/05/go-lib/pq-or-pgx-which-performs-better/ 
It is more reliable because pgx managing connection pool by itself and keep connections active.
Also I made some performance improvements. It's not just replacing of driver.

asim:
Ok that makes sense. Since its still postgres related could you please put it in postgres/pgx and then update the readme in the pgx dir itself to explain what it is.

plugin/postgres/pgx/db.go
```diff
@@ -0,0 +1,8 @@
+package pgxplugin
```
asim, line 1: Can you rename this to pgx rather than pgxplugin so it matches the plugin name

plugin/postgres/pgxplugin/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/micro/micro/plugin/pgx/v3
+module github.com/micro/micro/plugin/pgxplugin/v3
```
asim, line 1: I meant could you name it pgx not pgxplugin. Sorry for the confusion

Alexfilus: Same name as driver?

asim: Yea just "pgx" for the module and package name

PR #1966, opened 2022-09-07
Add darwin-arm64 as an supported platform

Joran-Dob:
Added darwin-arm64 as an supported platform the make the install script work as intended on Apple Silicon Macs.

This will make sure the 'Get Started' guide is also easily followable on Apple Silicon Macs. 
[https://github.com/micro/micro#install-binaries](https://github.com/micro/micro#install-binaries)

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -41,7 +41,7 @@ getSystemInfo() {
 }
 
 checkSupported() {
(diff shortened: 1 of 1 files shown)


asim:
Thanks. Are you a micro user?

Joran-Dob:
Not yet, but looking into it! Seems really cool, great work 😌

Issue #2086, opened 2025-03-22
V5 Clean Slate


asim:
Micro v1 - api, cli, web tooling around Go Micro
Micro v2 - transition to default of grpc and nats
Micro v3 - an end to end platform as a service
Micro v4 - vanity url / stripping bloated plugins

With V5 There's a feeling that a clean slate approach needs to be taken and to either re-merge with Go Micro or do something simpler. Personally I feel like a quickstart wrapper around Go Micro is quite useful, with some handy CLI tooling, runner, etc. So more like a dumbed down V3 that still offers the ability to use Go Micro plugins and interfaces. 

An alternative approach is to think a little more high level and create something purely designed around a protocol rather than a platform. A simple daemon that handles the networking aspects more broadly and conversion of JSON to gRPC.

Here's a starting point https://github.com/micro/micro/tree/v5-clean-slate. Ultimately Go Micro is still at the heart of development. Micro provides the CLI and additionally maybe some API and networking integration with things like tailscale.

asim:
V5 clean slate branch now merged. The work has begun.

PR #2090, opened 2025-04-23
V5 clean slate

asim:
A clean slate v5 that brings Go Micro back

https://github.com/micro/micro/issues/2086

- List services
- Call services
- Describe service
- Run API gateway
- Adds MCP server
- Serve via tailscale


asim:
@lambdaR ^^

asim:
@bketelsen any thoughts? ^^

bketelsen:
> @bketelsen any thoughts? ^^

ohh nice.  I'll pull and take a look later!

asim:
Now includes MCP server

bketelsen:
If I were to do an OS, I'd take this route: https://github.com/lxc/incus-os
As modern as an OS could possibly be with all the security features, immutable, can only interact with it via API.

Which would make really good sense in the Micro world

bketelsen:
overall this changeset fixes all of my issues and confusion with the state of Micro.

asim:
> If I were to do an OS, I'd take this route: https://github.com/lxc/incus-os As modern as an OS could possibly be with all the security features, immutable, can only interact with it via API.
> 
> Which would make really good sense in the Micro world

Ah very interesting. I can see that's still a Linux style OS where I feel like a more niche Go OS would just run Go Micro services as self contained binaries and expose them via the API. But yes this idea of purely connecting via an API makes sense. I'm not a gRPC fan these days. A clean http API with a CLI and SDK makes sense.

In our case the API is a very slim layer. I think the elusive part was being able to run a self contained system that mimicked an OS. Wasm gets us closer but without a true POSIX compliant API integration is very hard. Whatever we do, simple is better than clever.

